PR #31, opened 2022-11-01
Integrate chat system in application

antho1404:
This adds the chat plugin developed by Liteflow to the template so user can chat with each other

- [x] Integration of the chat window in app
- [x] Add state to preserve user's preferences
- [x] Resolve name based on app's data

### DIFF
--- a/components/ChatWindow.tsx
+++ b/components/ChatWindow.tsx
@@ -1,0 +1,96 @@
+import { Box, IconButton, useTheme } from '@chakra-ui/react'
+import type { Account } from '@nft/chat'
+import { Chat as ChatComponent, ChatProvider, useChat } from '@nft/chat'
+import { useSigner } from '@nft/hooks'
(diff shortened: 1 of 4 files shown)


EmmanuelDrouin (approved):
I'm good with this first implementation. We will need to think about mobile version as right now on mobile the widget is not properly display and takes a lot of spaces

achampagnedev:
Small comments: 

Before typing an address, cursor is very small, when entering at least 1 character the cursor is fine:
image

When creating a new chat and typing the address, the error message appears under the viewport. Maybe a solution is to pop the chat open when clicking on new.

image

Other design comment, The black border is a bid strong, maybe a gray border with drop shadow would look better, just opinion :).

Let me know if I need to create issues for these.

achampagnedev (approved):

EmmanuelDrouin:
[Messaging system - Marketplace Template implementation](https://trello.com/c/aaws6OHf/282-messaging-system-marketplace-template-implementation)

PR #43, opened 2022-11-27
Use theme directly rather than hook

antho1404:
Using the hook creates issues when the context is not created yet.
We already have the theme so that makes things much easier


cpvdeveloper (approved):
👍

PR #44, opened 2022-12-03
Update library with removal of session and delegation of providers to app

antho1404:
Dependency: https://github.com/liteflow-labs/liteflow-js/pull/90
Related: https://github.com/liteflow-labs/marketplace-template/issues/46

This dependency is easier to review in parallel with https://github.com/liteflow-labs/liteflow-js/pull/90 and follow the commit history.

This uses the latest version of the library that removes the session and most of the providers in the general context.

Thanks to that, the application can now define the web3 provider/connectors, the way to store the authenticated user (session / cookies), handle the user's authentication...

The application is now also responsible to reconnect the user's wallet with the eagerConnect still with the goal that it become the application's responsibility to handle the wallet of its choice


hooks/useEagerConnect.ts
```diff
@@ -0,0 +1,67 @@
+import { currentAccount } from '@nft/components/dist/session'
+import { useWeb3React } from '@web3-react/core'
+import { useCallback, useEffect, useState } from 'react'
+import { useCookies } from 'react-cookie'
+import { isMobile } from 'react-device-detect'
+import connectors from '../connectors'
+
+// Uniswap eager connect: https://github.com/Uniswap/uniswap-interface/blob/main/src/hooks/web3.ts#L15
```
cpvdeveloper, line 8: I think this link is outdated. Should it be pointing to this instead?
```
https://github.com/Uniswap/interface/blob/main/src/hooks/useEagerlyConnect.ts
```

hooks/useEagerConnect.ts
```diff
@@ -0,0 +1,67 @@
...
+        console.error('Failed to activate email connector', e)
+      }
+    }
+    setTried(true)
+  }, [activate, connectedAddress])
+
+  useEffect(() => {
+    if (active) return
+    void initialize()
+  }, [initialize]) /* eslint-disable-line react-hooks/exhaustive-deps */ // we need to re-run this on every update of the active connection, only when the initialized changes (and so the connectors)
```
cpvdeveloper, line 57: I don't understanding the comment you've left here, about when we **do** and **don't** want to run this `useEffect`.

Should it say "we **don't** need to re-run this on every..."?

hooks/useEagerConnect.ts
antho1404: Hum, that was the link already in the hooks, but it is not valid anymore. The hook you link seems to be the same but with a completely different logic. I would simply remove the comment knowing that anyway this part with integration of other wallets (and/or updating the lib that we have) will disappear

hooks/useEagerConnect.ts
antho1404: done

hooks/useEagerConnect.ts
antho1404: done

antho1404:
I updated the branch to merge into [feature/app-template-merge](https://github.com/liteflow-labs/marketplace-template/tree/feature/app-template-merge) as the build is failing because of some modification that requires to re-publish the library of templates but in the next PR the library or template will be completely gone so we can have a final full review on the [feature/app-template-merge](https://github.com/liteflow-labs/marketplace-template/tree/feature/app-template-merge) branch when ready

cpvdeveloper (approved):
✅

PR #48, opened 2022-12-09
Merge @nft/components into the component directory

antho1404:
Move all the components from `@nft/components` to the `./components` directory.
I also merged the translations and the theme.
The providers bugsnag and chakra have been added to the _app.
I also removed the option on the footer for the powered by present by default and application can simply remove it if they don't want it

While moving the components, I fixed a lot of errors from the linter:
- Add `key` in some components (mostly translations where we pass an array with the component)
- Remove the `children={<xxx>}` props to go with classic children. Mostly on the `InputRightElement` components
- Move query from navbar to layouts
- Rename some components that had the wrong name, `AssetCard` instead of `TokenCard`, for example

Note: the build is failing because I still need to update the hook library
Note 2: this PR is merged into https://github.com/liteflow-labs/marketplace-template/pull/47 so can safely be merged


codegen.yml
```diff
@@ -1,7 +1,7 @@
 schema: './node_modules/@nft/api-graphql/schema.graphql'
 documents:
   - './pages/**/*.gql'
-  - './pages/**/!(*.d).{ts,tsx}'
+  - './layouts/**/*.gql'
```
cpvdeveloper, line 4: Just want to check - have you intentionally not included the `/components` directory here? To encourage not writing graphql queries/mutations at the component level? I could understand this reasoning, but I can also see this being difficult to debug if someone does write a query/mutation collated with a component?

antho1404: Yes that's the idea, components should be pretty dumb and should not need to have graphql logic. If they do then it should be delegated maybe to a hook or passed as props.
Currently still the navbar is having queries but this is a bad design and this should be passed as props or the navigation should directly be merged into the layout

achampagnedev:
Should we add `es-mx` translations? I see we are missing some of the files/content.

achampagnedev:
@antho1404 Nevermind, got confused by seeing templates/components file added/modified to the other languages and es-mx having only one file, All good my bad!

ismailToyran (approved):
- Checked `.gql` files to make sure `codegen.yml` file covers all
- Compared every single file with the original ones from https://github.com/liteflow-labs/liteflow-js/pull/94/files#diff-4530c6b185cc9204d7821757322b5e37569652b0af89b5be0dc980a030f92713
- There are some renaming of the components
- There are removal of `children` prop from the `Input`'s
- There are added keys to translation files
- There is one additional key fix

Overall looking great, and every single change mentioned on the description seems to be great 👍 
The only thing I didn't check is translation json files.

cpvdeveloper (approved):
Running locally and all good ✅

PR #91, opened 2022-12-14
Added tabs on explore page, and added users page

achampagnedev:
Trello: https://trello.com/c/has1mFRK

Adds tabs to explore page, puts NFTs tab under `/explore` and Users tab under `/explore/users`

We are already using `/users` for profile. maybe we should switch to `/profile`, just a suggestion.


achampagnedev:
[Enhance user discovery and search - Explore add NFTs/Users tabs](https://trello.com/c/has1mFRK/434-enhance-user-discovery-and-search-explore-add-nfts-users-tabs)

EmmanuelDrouin (commented):
Looking good a few feedback:
- The card itself should be clickable and redirects to the user profile
  - Right now only the banner on the card is clickable
- There is an inconsistency with the spacing between the tabs compared the the spacing used on the user profile (which use the tab component)
  - On Explore we have spacing between tabs, but not on the user profile
  - Let's add 16px spacing between tabs on this component so it applies everywhere the component is use (same spacing for consistency)
  - Screen Shot 2565-12-19 at 10 37 55
  - Screen Shot 2565-12-19 at 10 38 09

achampagnedev:
@EmmanuelDrouin Made the adjustments from your feedback.

EmmanuelDrouin (approved):

antho1404:
I am just dropping by quickly. Keep in mind that the search and explore are on the same page, so we need to ensure that the search also works. 
From the home page, do a search for a user, then click on the tab user to see the results.
Expected: we should have the users with only the one matching our search

achampagnedev:
yes @antho1404 It was discussed today during the prod planning so it will be modified when implementing search

achampagnedev:
As just discussed with Antho this PR is ok for review as we found the right solution to implement search later one without modifying this structure @ismailToyran Give this PR a review when you get a chance

antho1404:
As this will break some UX on the platform until the search is done, let's merge it into a different branch so the main stays valid. Can you create a new branch, `features/explore-search-users`, and merge this PR into `features/explore-search-users` and then when the search PR is coming, merge that PR to `features/explore-search-users` as well, then later merge `features/explore-search-users` into `main`

achampagnedev:
New branch has been created

convert.ts
```diff
@@ -99,6 +99,20 @@ export const convertUser = (
   verified: user?.verification?.status === 'VALIDATED',
 })
 
+export const convertUserCard = (
```
antho1404, line 102: Usually, you would want to name this related to the data. Here `userWithCover` for example and not with the component that will be using this (as this might be multiple components) but not a big deal

pages/explore/explore.tsx
```diff
@@ -473,8 +477,33 @@ const ExplorePage: NextPage<Props> = ({
...
+          </Link>
+          <Link href="/explore/users" whiteSpace="nowrap">
+            <Tab as="div" borderColor="gray.200" pb={4} color="gray.500">
+              <Text as="span" variant="subtitle1">
+                {t('explore.tabs.users')}
+              </Text>
+            </Tab>
+          </Link>
+        </TabList>
+      </Tabs>
```
antho1404, line 503: Could be good to have a component about that as we have the same in the user but we talked about that and I guess that might come in another PR

pages/explore/users.tsx
```diff
@@ -0,0 +1,173 @@
...
+const UsersPage: NextPage<Props> = ({ offset, limit, page }) => {
+  const ready = useEagerConnect()
+  const { t } = useTranslation('templates')
+  const { data, refetch } = useFetchExploreUsersQuery({
+    variables: {
+      limit,
+      offset,
+    },
+  })
+  useExecuteOnAccountChange(refetch, ready)
```
antho1404, line 81:
```suggestion
```

This part is not necessary as the display doesn't depend on the connected user like the NFT where we can have unlocked content.

antho1404: fixed in [be047e4](https://github.com/liteflow-labs/starter-kit/pull/91/commits/be047e46760fb30538fb9fb3749e6a7bf25e9ba2)

convert.ts
antho1404: fixed in [4fbbdd3](https://github.com/liteflow-labs/starter-kit/pull/91/commits/4fbbdd3e4cf39b6e5387970f915c22fa3e1218c7)

pages/explore/explore.tsx
antho1404: fixed in the next PR

PR #97, opened 2022-12-21
Add search on users

achampagnedev:
Trello: https://trello.com/c/CfsK467E
Dependency: https://github.com/liteflow-labs/starter-kit/pull/91 https://github.com/liteflow-labs/starter-kit/pull/95

- Now able to search users
- Added a layout for the explore page

### DIFF
--- a/components/Explore.tsx
+++ b/components/Explore.tsx
@@ -1,7 +1,6 @@
 import {
   Flex,
   Heading,
-  Link,
   Spinner,
   Tab,
   TabList,
@@ -10,8 +9,8 @@ import {
 } from '@chakra-ui/react'
 import LargeLayout from 'layouts/large'
 import useTranslation from 'next-translate/useTranslation'
-import { useRouter } from 'next/router'
-import { FC, useCallback } from 'react'
+import { FC } from 'react'
+import Link from './Link/Link'
 
 const ExploreTemplate: FC<{
   title: string
@@ -21,22 +20,7 @@ const ExploreTemplate: FC<{
   children: JSX.Element
 }> = ({ title, loading, search, selectedTabIndex, children }) => {
   const { t } = useTranslation('templates')
-  const { replace } = useRouter()
-
-  const handleTabClick = useCallback(
-    async (tabHref: string) => {
-      if (search) {
-        return await replace({
-          pathname: tabHref,
-          query: { search },
-        })
-      }
-      return await replace({
-        pathname: tabHref,
-      })
-    },
-    [replace, search],
-  )
+  const searchParam = search ? `?search=${search}` : ''
 
   return (
     <LargeLayout>
@@ -56,11 +40,7 @@ const ExploreTemplate: FC<{
         overflowX="auto"
       >
         <TabList>
-          <Link
-            onClick={() => handleTabClick('/explore')}
-            whiteSpace="nowrap"
-            mr={4}
-          >
+          <Link href={`/explore${searchParam}`} whiteSpace="nowrap" mr={4}>
             <Tab as="div" borderColor="gray.200" pb={4} color="gray.500">
               <Text as="span" variant="subtitle1">
                 {t('explore.tabs.nfts')}
@@ -68,8 +48,9 @@ const ExploreTemplate: FC<{
             </Tab>
           </Link>
           <Link
-            onClick={() => handleTabClick('/explore/users')}
+            href={`/explore/users${searchParam}`}
             whiteSpace="nowrap"
+            mr={4}
           >
             <Tab as="div" borderColor="gray.200" pb={4} color="gray.500">
               <Text as="span" variant="subtitle1">
--- a/components/Explore.tsx
+++ b/components/Explore.tsx
@@ -17,9 +17,9 @@ const ExploreTemplate: FC<{
   title: string
   loading: boolean
   search: string | null
-  tabIndex: number
+  selectedTabIndex: number
   children: JSX.Element
-}> = ({ title, loading, search, tabIndex, children }) => {
+}> = ({ title, loading, search, selectedTabIndex, children }) => {
   const { t } = useTranslation('templates')
 
   return (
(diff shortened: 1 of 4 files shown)


components/Explore.tsx
```diff
@@ -0,0 +1,90 @@
...
+      return await replace({
+        pathname: tabHref,
+      })
+    },
+    [replace, search],
+  )
+
+  return (
+    <LargeLayout>
+      <Head title={title} />
```
antho1404, line 44: Let's keep this in the page so we have only one place to find `Head`

components/Explore.tsx
```diff
@@ -0,0 +1,90 @@
...
+  title: string
+  loading: boolean
+  search: string | null
+  selectedTabIndex: number
+  children: JSX.Element
+}> = ({ title, loading, search, selectedTabIndex, children }) => {
+  const { t } = useTranslation('templates')
+  const { replace } = useRouter()
+
+  const handleTabClick = useCallback(
```
antho1404, line 27: I'm not sure to understand the reason for going with the `replace` method. If we already have the `search` then we can simply build the URL and pass a single href to the tabs instead of the `onClick`.

pages/explore/users.tsx
```diff
@@ -1,40 +1,35 @@
...
 type Props = {
   limit: number
   page: number
   offset: number
+  queryFilter: AccountFilter[]
+  search: string | null
 }
+const searchFilter = (search: string): AccountFilter =>
+  ({
+    or: [{ name: { includesInsensitive: search } } as AccountFilter],
```
antho1404, line 31: Let's also search by address, I think that will be super useful, that's how I search most of the time on the different platforms

pages/explore/users.tsx
```diff
@@ -50,10 +45,17 @@ export const getServerSideProps = wrapServerSideProps<Props>(
         : parseInt(ctx.query.page, 10)
       : 1
     const offset = (page - 1) * limit
+    const search =
+      ctx.query.search && !Array.isArray(ctx.query.search)
+        ? ctx.query.search
+        : null
```
antho1404, line 51: Why this pattern? All the others params are using something different. If we have 2 search in the query then none will be read while the pattern use for `limit` or `page` takes the first one

components/Explore.tsx
achampagnedev: It was creating a page reload when just passing an href, do you have an idea why ?

pages/explore/users.tsx
achampagnedev: this was copied from explore.tsx

antho1404: ok indeed then let's keep it like that then

components/Explore.tsx
antho1404: Most likely because you need to use the `next/link` in the tabs or in this case I believe we have a `Link` component that uses the chakra + next link so I guess if you use the `/components/Link` that should work without any refresh

achampagnedev: That's what I did early on and had the refresh issue, discussed it with @NicolasMahe and referred me to do what we did on the explore page when submitting filters.

pages/explore/users.tsx
achampagnedev: added

components/Explore.tsx
achampagnedev: moved back to pages.

components/Explore.tsx
antho1404: I think that was unrelated, I tried and the link works perfectly with no refresh with the following

```diff
diff --git a/components/Explore.tsx b/components/Explore.tsx
index d07191a..212e113 100644
--- a/components/Explore.tsx
+++ b/components/Explore.tsx
@@ -1,7 +1,6 @@
 import {
   Flex,
   Heading,
-  Link,
   Spinner,
   Tab,
   TabList,
@@ -10,8 +9,8 @@ import {
 } from '@chakra-ui/react'
 import LargeLayout from 'layouts/large'
 import useTranslation from 'next-translate/useTranslation'
-import { useRouter } from 'next/router'
-import { FC, useCallback } from 'react'
+import { FC } from 'react'
+import Link from './Link/Link'
 
 const ExploreTemplate: FC<{
   title: string
@@ -21,22 +20,8 @@ const ExploreTemplate: FC<{
   children: JSX.Element
 }> = ({ title, loading, search, selectedTabIndex, children }) => {
   const { t } = useTranslation('templates')
-  const { replace } = useRouter()
 
-  const handleTabClick = useCallback(
-    async (tabHref: string) => {
-      if (search) {
-        return await replace({
-          pathname: tabHref,
-          query: { search },
-        })
-      }
-      return await replace({
-        pathname: tabHref,
-      })
-    },
-    [replace, search],
-  )
+  const searchParam = search ? `?search=${search}` : ''
 
   return (
     <LargeLayout>
@@ -56,21 +41,14 @@ const ExploreTemplate: FC<{
         overflowX="auto"
       >
         <TabList>
-          <Link
-            onClick={() => handleTabClick('/explore')}
-            whiteSpace="nowrap"
-            mr={4}
-          >
+          <Link href={`/explore${searchParam}`} whiteSpace="nowrap" mr={4}>
             <Tab as="div" borderColor="gray.200" pb={4} color="gray.500">
               <Text as="span" variant="subtitle1">
                 {t('explore.tabs.nfts')}
               </Text>
             </Tab>
           </Link>
-          <Link
-            onClick={() => handleTabClick('/explore/users')}
-            whiteSpace="nowrap"
-          >
+          <Link href={`/explore/users${searchParam}`} whiteSpace="nowrap">
             <Tab as="div" borderColor="gray.200" pb={4} color="gray.500">
               <Text as="span" variant="subtitle1">
                 {t('explore.tabs.users')}
```


https://user-images.githubusercontent.com/1783126/209292126-d29d451f-a3e5-48c4-9c85-9f5c0d737e59.mov

achampagnedev:
@antho1404 Yes that worked. an other question maybe more for @EmmanuelDrouin What do you think of adding a check when search if we are already on the user's page ? for example if I search a user, I might want to search a user again next. Right now when searching it will always go back to the NFT page, we could add a check to verify if we are on the users page or not and stay on the users page when we search again if we are.

antho1404:
Definitely a good thing to add, I created another issue to not block this PR https://github.com/liteflow-labs/starter-kit/issues/108

antho1404:
@achampagnedev please don't forget to add proper titles to your PRs

antho1404 (approved):

PR #116, opened 2023-01-04
New Token Card UI

achampagnedev:
### Project organization

- Trello card: https://trello.com/c/l1ZLMDa3
- Related https://trello.com/c/FV5FMleK/43-collection-work

### Description

Implements the new [design](https://www.figma.com/file/RIHKWJvqf93HOpF4zxifvf/LTF-%E2%80%94-Templates?node-id=245%3A26424&t=HS3yfbN8ivfpRe7R-0) for the Token Cards 

### How to test

Token cards are visible on multiple sections of the application:
- Home page
- Explore page
- Profile pages
- Creation page

Make sure that you have all different kinds of cards:
- Cards to NFT you own
- Cards with Direct sale
- Cards with timed auction


components/Sales/Auction/CardFooter.tsx
```diff
@@ -1,6 +1,6 @@
-import { Avatar, Flex, HStack, Icon, Tag, Text } from '@chakra-ui/react'
+import { Flex, HStack, Icon, Stack, Text } from '@chakra-ui/react'
```
antho1404, line 1: Here I am connected as another user in the profile page and I have the wrong footer. I should have something about placing a bid no?
Screenshot 2023-01-06 at 14 35 47

components/Sales/Direct/CardFooter.tsx
```diff
@@ -1,4 +1,4 @@
-import { Flex, HStack, Tag, TagLabel, Text } from '@chakra-ui/react'
+import { Flex, HStack, Stack, Text } from '@chakra-ui/react'
```
antho1404, line 1: Similar on the profile I have "view my NFT" even when I'm disconnected

components/Sales/Open/CardFooter.tsx
```diff
@@ -1,44 +1,42 @@
-import { Flex, Tag, Text } from '@chakra-ui/react'
+import { Flex, Stack, Text } from '@chakra-ui/react'
```
antho1404, line 1: Similar issue on profile

antho1404 (commented):
I didn't finish the review but already a few feedback to fix. You can already work on some fix for that and in the meantime I will continue my review/tests

achampagnedev:
[Collection work - Update token card UI](https://trello.com/c/l1ZLMDa3/427-collection-work-update-token-card-ui)

achampagnedev:
@antho1404 The issue about showing the proper ownership on the profile pages should be resolved now

achampagnedev:
Some possible feedback when placing a bid when seeing the card on the left column: 
image

Do we want a way to not display the report button, or force the hover state to true? let me know your opinion

antho1404:
I would personally remove this menu in the card as it does not provide a lot of value and having a quick access to report doesn't provide much value. If we had other actions like "sale", "place bid", "gift", "add to watchlist"... I would say why not for this menu but just for a report I don't think we should have a menu as this should not be prominent

EmmanuelDrouin:
> I would personally remove this menu in the card as it does not provide a lot of value and having a quick access to report doesn't provide much value. If we had other actions like "sale", "place bid", "gift", "add to watchlist"... I would say why not for this menu but just for a report I don't think we should have a menu as this should not be prominent

The idea of this menu is to open the door for additional actions as you described. For now, we only have the report (current action on the app) but we could add more useful shortcuts here that will improve the UX and interaction from users. I can come up with few additional and useful actions here if you want to push this further now ;)

So I want to keep this menu here to push the starter kit forward and start having based functionality that users might expect. If you want we can add right now additional shortcuts (which will take a bit more time) but Not adding this menu because we don't add additional functionality yet (to save time) is not a good direction from my opinion.

Regarding your question @achampagnedev
> Do we want a way to not display the report button, or force the hover state to true? let me know your opinion
Yes please disable that and remove the button with over state (can do a separate component if needed), this card should not be the exact same card component with the same functionality as this doesn't make sense in the context of the form.

EG 1: if you display the hover button with the place bid action, what's the point, this will redirect to the same page you already are. 
EG 2: There is no point in providing additional actions (cf ellipsis menu) if you are in a form as in this case the user is focused on purchasing/listing/or making an offer on the item, the user is not trying to achieve any other actions.

EmmanuelDrouin (commented):
Just started my review but I saw right away that there is a wrong logic implementation on the card compared to the original design, let me know if this is because you think the original design logic was not great or if this is a misunderstanding.

**Current implementation:**
- The bottom of the card doesn't display the Bid info if you don't hover the card
- Tho the logic for displaying action button doesn't work, sometime we display a primary action button while hovering the card sometimes not (eg: auction)
  - Screen Shot 2566-01-10 at 12 37 34
  - Screen Shot 2566-01-10 at 12 37 41
  - Screen Shot 2566-01-10 at 12 37 53

**Design logic that was supposed to be implemented:**
- By default we always display at the bottom of the card the bid information related to the item (gray section)
  - This allows us to display the users' offers related to the item all the time, which is useful for auction and open offers, but also for fixed price listing if a user wants to compare quickly the listing price vs the offers received
- While hovering the card, the primary action button is spossued to overlaps this "bid section" similar to looksrare
- Design file > [Card component](https://www.figma.com/file/RIHKWJvqf93HOpF4zxifvf/LTF-%E2%80%94-Templates?node-id=245%3A26424&t=QMik3j98Mz6TQCl5-4)

If you think my logic is not good, i'm open to quickly brainstorm a better solution, but this should be address now before moving forward as the current system implemented doesn't fully work for me, so please let me know @antho1404 @achampagnedev

EmmanuelDrouin:
Following a brainstorming session with Alex to improve the overall logic of the new card design, the component has been updated accordingly on Figma > https://www.figma.com/file/RIHKWJvqf93HOpF4zxifvf/LTF-%E2%80%94-Templates?node-id=245%3A26424&t=QMik3j98Mz6TQCl5-4

These changes simplify the card structure, provide more consistency in terms of data positioning and logic and should improve the overall UX of this component ;)

achampagnedev:
@EmmanuelDrouin Just updated the Token's UI according to latest changes, let me know what you think

EmmanuelDrouin (commented):
My feedback:
- The ellipsis button to display the dropdown menu is supposed to be hidden by default and appear while hovering the card
Screen Shot 2566-01-12 at 11 53 26
- The redirection on click on the card primary button (place a bid/buy now) are redirecting to the token detail page. Please update the redirection to redirect to the page related (place a bid/checkout)
- Auction card, if a bid has been placed the bottom of the card still display the label "Ongoing auction" when it supposed to display the highest bid placed (same logic as the open bid cards)
- Same comment for "open for bids" card, the logic doesn't work when a bid is placed the bottom of the card still display the label "open for bids" where it should display the highest bid placed

achampagnedev:
@EmmanuelDrouin I just pushed an update for the 3 feedbacks you just provided

achampagnedev:
@antho1404 @EmmanuelDrouin If you can do the final review I think this is now good to go.

antho1404:
One thing to keep in mind with this PR, let's not merge it until we have the collections as it adds a new link to the collection pages that does not exists yet

EmmanuelDrouin:
@achampagnedev last feedback, all the other one have been fixed ;)

When it's my NFT, on lick on "view my NFT" I got redirected to a 404 
- 404 > https://nft-test-bsc-testnet-pqbp2p6ea-liteflow.vercel.app/tokens/97-0xcfbb45ff528c8ff6542887f631a1f704b92cf7db-60249402084937876423066029128237587855293854847399126863749031985230051891574false
- Token I'm trying to access > https://nft-test-bsc-testnet-pqbp2p6ea-liteflow.vercel.app/tokens/97-0xcfbb45ff528c8ff6542887f631a1f704b92cf7db-60249402084937876423066029128237587855293854847399126863749031985230051891574

achampagnedev:
@EmmanuelDrouin introduced issue should be fixed, feel free to test and approve is that's ok for you :).

EmmanuelDrouin:
@achampagnedev deployment issue let me know once fixed so I can do a final validation ;)

achampagnedev:
@EmmanuelDrouin Fixed the deploy.

EmmanuelDrouin:
@antho1404 could you do a final review on this PR please? I think this is good to go

components/Token/Card.tsx
```diff
@@ -46,6 +78,7 @@ export type Props = {
       }
     | undefined
   numberOfSales: number
+  displayCreator?: boolean
```
antho1404, line 81: Why this `displayCreator`? I don't think it's used anywhere

convert.ts
```diff
@@ -27,20 +27,75 @@ export const convertAsset = (
...
+      address: asset.collection.address,
+      name: asset.collection.name,
+    },
+    owned: BigNumber.from(asset.owned.aggregates?.sum?.quantity || '0'),
+    unlockedContent: asset.unlockedContent,
+    bestBid: {
+      unitPrice: BigNumber.from(bestBid.unitPrice),
+      currency: bestBid.currency,
+    },
+  }
```
antho1404, line 97: Instead of returning 2 type of object that can lead to confusion and mistakes
```suggestion
  return {
    id: asset.id,
    animationUrl: asset.animationUrl,
    image: asset.image,
    name: asset.name,
    collection: {
      address: asset.collection.address,
      name: asset.collection.name,
    },
    owned: BigNumber.from(asset.owned.aggregates?.sum?.quantity || '0'),
    unlockedContent: asset.unlockedContent,
    bestBid: bestBid ? {
      unitPrice: BigNumber.from(bestBid.unitPrice),
      currency: bestBid.currency,
    } : undefined,
  }
```

convert.ts
```diff
@@ -66,21 +116,55 @@ export const convertAssetWithSupplies = (
...
+      ),
+      totalSupply: BigNumber.from(
+        asset.ownerships.aggregates?.sum?.quantity || '0',
+      ),
+      owned: BigNumber.from(asset.owned.aggregates?.sum?.quantity || '0'),
+      bestBid: undefined,
+    }
+  }
+
+  return {
```
antho1404, line 144: Similar comment here

antho1404 (commented):
Pretty good, I believe also because we have the address of the current user in pretty much all queries we could remove all the `useExecuteOnAccountChange(refetch, ready)`.

A few feedback but otherwise really good. Let me know if you want I take care of these feedback

components/Token/Card.tsx
achampagnedev: @antho1404 It's not used anywhere but as in the specs, in the case of a collection focused on creator we need to be able to display the creator instead of the collection: 

image

components/Sales/Auction/CardFooter.tsx
```diff
@@ -1,15 +1,12 @@
-import { Avatar, Flex, HStack, Icon, Tag, Text } from '@chakra-ui/react'
+import { Flex, Text } from '@chakra-ui/react'
 import { BigNumber } from '@ethersproject/bignumber'
-import { HiOutlineClock } from '@react-icons/all-files/hi/HiOutlineClock'
+import Price from 'components/Price/Price'
```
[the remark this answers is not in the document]
achampagnedev: Yup seems I will have to slap my VSC's hand, this is from the auto import cmd + ,

achampagnedev:
Updated with the other change requests, see my remaining comments about `displayCreator`

components/Token/Card.tsx
antho1404: 👍

antho1404 (approved):

PR #120, opened 2023-01-07
Debug 500 errors on cold start

antho1404:
Way to reproduce the issue:
Push a commit with a new function then watch the logs, as simple as this, this first run of the function will trigger the error

Things I've tried:
- Force context as https://stackoverflow.com/questions/71545636/netlify-apollo-nextjs-ssr-client-network-socket-disconnected-before-secure-tls-c

Interesting things to look:
- never ending function: https://stackoverflow.com/questions/64885483/error-client-network-socket-disconnected-before-secure-tls-connection-was-esta
- Some stuff with proxy header https://stackoverflow.com/questions/53593182/client-network-socket-disconnected-before-secure-tls-connection-was-established


NicolasMahe:
I checked the functions logs on vercel and the different apps and I notice the error everywhere, except on:
- hodooi mainnet and testnet
- onlyshoot mainnet and testnet
They both use the @nft packages version `1.0.0-beta.8`.

For the admin, the issue is present on the demo and test admins but not on the clients admin.
The client's admins were deployed 83 days ago with the following commit: https://github.com/liteflow-labs/admin/commit/dffd86f4c8b73f093ad37f6bc6eaa4234eb0c178.
Also using @nft package version `1.0.0-beta.8`.

I think there is a bug introduce after the release `1.0.0-beta.8` of the libs.

NicolasMahe:
Looking at the error message “Client network socket disconnected before ….” it may be because of a query that is executed but not waited for.
I think the query “GetConfig” is causing the issue. It’s executed on the SSR function, which should not.
I think the trigger of this GetConfig query on SSR is because of a useMemo in the LiteflowProvider.
useMemo are executed on the first render (as opposed as useEffect) causing the execution of the GetConfig query that is immediately “discard” thus throwing a network error because the client already disconnect

antho1404:
Good spot, this was indeed the issue, I was searching the wrong way
Here is the PRs to fix this:
- https://github.com/liteflow-labs/starter-kit/pull/121
- https://github.com/liteflow-labs/liteflow-js/pull/120

Closing this PR to keep git clean

PR #125, opened 2023-01-10
Adds display of rarity on the Token Detail page

achampagnedev:
### Project organization

- Trello card: https://trello.com/c/jxxAOZOY

### Description

Adds rarity to the properties on the Token Detail page

### How to test

visit a Token Detail Page, see if properties have the rarity.


achampagnedev:
[Collection work - Add rarity to properties](https://trello.com/c/jxxAOZOY/433-collection-work-add-rarity-to-properties)

components/Trait/TraitList.tsx
```diff
@@ -27,7 +32,9 @@ const TraitList: FC<TraitListProps> = ({ traits }) => {
             isTruncated
             pb={1}
           >
-            {t(`categories.${trait.type}`, null, { fallback: trait.type })}
+            {trait.type !== 'Category'
```
antho1404, line 35: I'm not sure this condition is right I guess it's the opposite no

antho1404 (changes requested):
The calculation is wrong, I have 0.17% which is extremely low where it should most likely be 17%.
Also make sure to pick a collection on the explore then filter through the different categories and the sum of all the rarity should be 100% otherwise there is a problem somewhere.

achampagnedev:
@antho1404 Updated the calculation so it should be in the right range now. By going through the categories for ERC721 I am getting these results: 

```Art
76.67% rarity

DeFi
3.33% rarity

Photography
16.67% rarity```

Seems like we are missing a 3.33%. Looking at the calculations we're doing though it seems right:
<img width="644" alt="image" src="https://user-images.githubusercontent.com/10590599/211728393-84b5c62f-de65-436e-a64d-b24deb1102b6.png">

This was taken from O-Mee but has been slighting modified. The calculation should still be the same as on O-Mee though.

**Comment by antho1404** [comment]
Yes here are my result as well across all the demos:

```
demo-bsc-testnet 1155: 94.87%
demo-bsc-testnet 721: 99.99% (might be an issue on my side)

nft-test-bsc-testnet 1155: 100%
nft-test-bsc-testnet 721: 96.67%

nft-test-polygon-mumbai 1155: 58.82%
nft-test-polygon-mumbai 721: 62.79%
```

We do have some valid results. My guess is that we have an issue with the API that calculates this. I created an issue about it https://github.com/liteflow-labs/backend/issues/2428

**Comment by antho1404** [comment]
One thing still doesn't really make sense to me with this dev.
We are displaying the percentage of assets having this specific trait (which is good and what most marketplaces have), but then we are saying that this is the "percentage rarity" which doesn't make much sense.
It could make sense if the rarest would be the highest percentage of rarity, but this is not the case as an asset with a trait only to this asset will have something like 1% then now we display "1% rarity" which definitely feels like it's not rare.

What's the rationale behind this? Should we remove the "rarity" or should we inverse the percentage so 1% would become 99% rarity? (I personally would recommend removing the "rarity" cc @EmmanuelDrouin

**Comment by NicolasMahe** [comment]
> Also make sure to pick a collection on the explore then filter through the different categories and the sum of all the rarity should be 100% otherwise there is a problem somewhere.

The backend doesn't enforce Asset to have Categories in Trait. There may be some Assets without Categories, especially on the test backends.

Using the explore page, you can see we have different count when filtering with a specific collection:
https://demo-bsc-testnet-e5xfe5b4h-liteflow.vercel.app/explore?collections=97-0xaade84b6a5f76c0a5f7ae91e65197329eb23763f

And filtering the same collection but with all the categories:
https://demo-bsc-testnet-e5xfe5b4h-liteflow.vercel.app/explore?collections=97-0xaade84b6a5f76c0a5f7ae91e65197329eb23763f&categories=Art&categories=Currency&categories=DeFi&categories=Games&categories=Photography

**Comment by EmmanuelDrouin** [comment]
> What's the rationale behind this? Should we remove the "rarity" or should we inverse the percentage so 1% would become 99% rarity? (I personally would recommend removing the "rarity"

We were following Rarible on this design, this is the wording they are using. We could either remove it totally and keep only the percentage, or add a label next to the percentage like Gem "have this trait" up to you guys

**Comment by antho1404** [comment]
Rarible is different because they display a table, so the percentage is not directly related to the rarity, it's a column about the rarity, and they display a percentage without saying the actual "unit" of this number (like you would have a table showing an attribute "length" for example that would show 32 or 32cm).
Here the "unit" is the "own this trait" and the type is the rarity, but the unit is not the rarity (like I wouldn't say "32 length" in my example).

File: convert.ts [hunk 2]
```diff
@@ -80,6 +82,40 @@ export const convertAssetWithSupplies = (
...
+      ({ type }) => type === assetTrait.type,
+    )
+    const traitValueCount =
+      traitInCollection?.values?.find(({ value }) => value === assetTrait.value)
+        ?.count || 0
+    return {
+      type: assetTrait.type,
+      value: assetTrait.value,
+      totalCount: traitValueCount,
+      rarity: (traitValueCount / asset.collection.supply) * 100,
```
NicolasMahe, line 112: Following conversation during the meeting, let's rename this prop to something else to reduce confusion with an actual rarity index.

antho1404 (approved):

PR #131, opened 2023-01-13
Explore collections and search the collections

achampagnedev:
### Project organization

- Trello card: https://trello.com/c/Ixhvxhzl
- Trello card: https://trello.com/c/TrE2jJuD

### Description

Adding the Collections tab and its search functionality. I should be working the same the Users tab. 
We should be able to search name, address and description.

### How to test

Visit the explore page. Click on collections. You should be able to see the listing of collections, we currently only have 2 collections. I discussed with @EmmanuelDrouin about having some placeholder images for our demo collections.

If you want to test the pagination/results per page you can duplicated the array of the collections in the codebase but this is the exact same code as the users page.


achampagnedev:
@EmmanuelDrouin Please notify me when we have some collection placeholder images so I can add them to the demo

achampagnedev:
[Collection work - Search functionality](https://trello.com/c/TrE2jJuD/431-collection-work-search-functionality)

achampagnedev:
[Collection work - Collection tab + card component](https://trello.com/c/Ixhvxhzl/430-collection-work-collection-tab-card-component)

achampagnedev:
@achampagnedev missing volume sorting

achampagnedev:
@achampagnedev I noticed when changing the sort we can see the "empty" state for a second. I looked at the explore/nfts page and this doesn't happen. Trying to find how to avoid this. Any feedback is welcome

EmmanuelDrouin:
@achampagnedev initial feedback

While on the explore page > Collections
- The sort-by is displayed in Japanese or Chinese (bug)
  -  Screen Shot 2566-01-16 at 17 20 48
- Could you explain the reason for putting the "all-time" volume as the first element in the dropdown? I'm not against the idea but I find it a bit weird in terms of orga as you will expect something like "24h > 7d > 14d > 28d > all-time" so having this at the end of the list

I don't actually know (think) if the volume display on the card is actually right, as no matter the sorting I'm choosing for the volume the data displayed in both cards don't change (total volume).
So it looks like it doesn't work as if I select the volume for the last 24h these data should be reflected on the card (total volume should change) compared to all-time. 
The "total volume" is based on the period of time selected, so like on O-MEE these change should be reflected on the card

achampagnedev:
@EmmanuelDrouin 

- Language bug: will fix 
- Default sort choice : All-time is the default selected according to the figma file (assumed it was the default). I am definitely okay with changing it to 24h as I feel it's more relevant.
- I will look on O-mee if the volume is updated but as of currently, the collection listin crashes if you try to change the card sizes

achampagnedev:
Language bug + default sort to 24h have been updated.

EmmanuelDrouin:
> Default sort choice : All-time is the default selected according to the figma file (assumed it was the default). I am definitely okay with changing it to 24h as I feel it's more relevant.

This is indeed intended to have "all-time" selected by default (as we don't have enough data/activity on a short period of time on testnet). My point was on the order inside the dropdown; currently "all-time" is displayed on top (on O-MEE it is at the bottom) so would like to know a bit more on this specific point ;)

> I will look on O-mee if the volume is updated but as of currently, the collection listin crashes if you try to change the card sizes

Yes we have this issue which is a bit annoying when we want to compare both platforms right now haha, hopefully, this will be fixed soon ;)

Regarding the placeholder for the open collections on the demo you can find the design here > [721 collection](https://www.figma.com/file/0AHxoPROqyuCllH6JqArVQ/Marketplace-Template---UI-Flows?node-id=1325%3A133304&t=7ktGwVAUFiCKVp3K-4) / [1155 collection](https://www.figma.com/file/0AHxoPROqyuCllH6JqArVQ/Marketplace-Template---UI-Flows?node-id=1326%3A105348&t=7ktGwVAUFiCKVp3K-4)
- For both descriptions, I used some Lorem Ipsum > https://www.lipsum.com/feed/html
- Regarding the "branding" of these collections, we should remove the "Liteflow branding" to follow what we did with the "Acme" redesign for the demo (to not drive the idea of our own platform)
  - Got an agreement from Anthony on this
  - So I assume we will do this update quickly (hopefully during this week) and tho implement the profile image and cover accordingly ;)

achampagnedev:
@EmmanuelDrouin After looking at the source code from O-Mee the sorting should be working the same on the starter kit. It would help to have proper collections imported to facilitate the current work on collections though

EmmanuelDrouin:
> default sort to 24h have been updated.

Just to clarify the default sort needs to be "all-time" as described on my previous comment, my question was on:
> My point was on the order inside the dropdown; currently "all-time" is displayed on top (on O-MEE it is at the bottom) so would like to know a bit more on this specific point ;)

FYI I can't access the "collections" tab on the explore page anymore on any deployment, it displays a blank page (even after hard refresh)

achampagnedev:
@EmmanuelDrouin After fixing the problem for the collections page it seems to come from a test collection that was added to starter kit: 

image

It has no floor price therefore it made the cards crash. I will push a fix in case there is no floor price.

antho1404 (approved):
Small feedback but otherwise good to go

achampagnedev:
@antho1404 Just pushed updates :)

Issue #134, opened 2023-01-19
Add min size to cards on some layouts


antho1404:
With the new card design, some pages displaying cards are broken

<img width="891" alt="Screenshot 2023-01-19 at 13 15 08" src="https://user-images.githubusercontent.com/1783126/213369271-96065db0-e788-4125-8c9d-e496fa44c5a6.png">
<img width="821" alt="Screenshot 2023-01-19 at 13 15 20" src="https://user-images.githubusercontent.com/1783126/213369301-f9802213-b719-44bc-bb84-b034118331df.png">

- Checkout page
- Bid page
- Creation page
- Maybe other

EmmanuelDrouin:
Closing this issue as #136 is merged

PR #136, opened 2023-01-19
Fix card preview on some pages

antho1404:
Fix https://github.com/liteflow-labs/starter-kit/issues/134

The card was broken in some places
- Place a bid
- Purchase
- Create an offer
- Create an nft

Also when and image had an error the display would break

This solves everything :)

I applied a grid of 1 third / 2 third for these pages to ensure that the card fits into a third of the page or the full width in mobile


EmmanuelDrouin (approved):
Perfect!

PR #139, opened 2023-01-23
Webhook email notification + bid created and auction bid created

antho1404:
### Project organization

- Dependency: https://github.com/liteflow-labs/starter-kit/pull/140

### Description

Add a new `/api/notification` function that can be added as a custom webhook.
This also adds new email support with https://react.email/ that are sent directly from this function.
I also added 2 emails
- `bid created` when a bid is created
- `auction bid created` when a bid is created from an auction

### How to test

Setup new required envs
- EMAIL_PASSWORD
- EMAIL_FROM
- LITEFLOW_WEBHOOK_SECRET
Use the new webhook feature on the infra and connect the following webhooks
- `BID_CREATED` to the `/api/notification`
- `AUCTION_BID_CREATED` to the `/api/notification`
Create a new open bid or auction bid and check that you received your email correctly


package.json
```diff
@@ -26,6 +26,17 @@
...
+    "@react-email/container": "^0.0.5",
+    "@react-email/head": "^0.0.3",
+    "@react-email/hr": "^0.0.3",
+    "@react-email/html": "^0.0.3",
+    "@react-email/img": "^0.0.3",
+    "@react-email/link": "^0.0.3",
+    "@react-email/preview": "^0.0.4",
+    "@react-email/render": "^0.0.5",
+    "@react-email/section": "^0.0.4",
+    "@react-email/text": "^0.0.3",
```
NicolasMahe, line 39: Let's keep it simple for this first version and use basic html template, exactly like we currently have in the backend.
We will improve the dev ux and design of the emails in the future.

pages/api/notification.tsx
```diff
@@ -0,0 +1,91 @@
...
+        to: payload.taker.email,
+        subject: `New bid received for ${payload.quantity} edition${
+          BigNumber.from(payload.quantity).gt(1) ? 's' : ''
+        } of ${payload.asset.name} for ${formatUnits(
+          payload.unitPrice.amount,
+          payload.unitPrice.currency.decimals,
+        )} ${payload.unitPrice.currency.symbol}${
+          BigNumber.from(payload.quantity).gt(1) ? ' each' : ''
+        }`,
+        html: render(<BidCreated {...data} />),
```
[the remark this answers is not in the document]
antho1404: Did a massive simplification on this function with a simple mapping. Still some edit needed if we want to change the type of email sent but not the email itself
```
  ['BID_CREATED', [BidCreated]],
  ['AUCTION_BID_CREATED', [AuctionBidCreated]],
```

package.json
antho1404: removed for now

.env.example
```diff
@@ -14,6 +14,9 @@ PINATA_GATEWAY="liteflow.mypinata.cloud"
 NEXT_PUBLIC_OFFER_VALIDITY_IN_SECONDS="2419200"
 NEXT_PUBLIC_AUCTION_VALIDITY_IN_SECONDS="604800"
 NEXT_PUBLIC_BASE_URL=https://demo.liteflow.com
+EMAIL_PASSWORD="SG."
+EMAIL_FROM="Liteflow <contact@liteflow.com>"
+LITEFLOW_SECRET=super_secret
```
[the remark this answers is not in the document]
antho1404: done

pages/api/notification.tsx
```diff
@@ -0,0 +1,91 @@
...
+import invariant from 'ts-invariant'
+import AuctionBidCreated from '../../emails/AuctionBidCreated'
+import BidCreated from '../../emails/BidCreated'
+
+const transporter = nodemailer.createTransport({
+  host: 'smtp.sendgrid.net',
+  port: 587,
+  auth: {
+    user: 'apikey',
+    pass: process.env.EMAIL_PASSWORD,
```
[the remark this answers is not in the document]
antho1404: I didn't want to add a complex config for the email, and simply the SendGrid key that could be our recommendation could be simpler and if they have the knowledge to choose their providers, then the probably have the knowledge to change the provider.
Anyway moved everything in env for now, we can simplify later if needed

pages/api/notification.ts
```diff
@@ -0,0 +1,57 @@
+import { Events, parseAndVerifyRequest } from '@nft/webhook'
+import { NextApiRequest, NextApiResponse } from 'next'
+import nodemailer, { SendMailOptions } from 'nodemailer'
+import invariant from 'ts-invariant'
+import AuctionBidCreated from '../../emails/AuctionBidCreated'
+import BidCreated from '../../emails/BidCreated'
+
+const transporter = nodemailer.createTransport({
```
NicolasMahe, line 8: Shouldn't we add checks to make sure the envs are properly set?
If they are not set, it try to send using localhost.

```suggestion
invariant(process.env.EMAIL_HOST, 'env EMAIL_HOST is required')
invariant(process.env.EMAIL_PORT, 'env EMAIL_PORT is required')
invariant(process.env.EMAIL_USERNAME, 'env EMAIL_USERNAME is required')
invariant(process.env.EMAIL_PASSWORD, 'env EMAIL_PASSWORD is required')

const transporter = nodemailer.createTransport({
```

pages/api/notification.ts
```diff
@@ -0,0 +1,57 @@
...
+  ['BID_CREATED', [BidCreated]],
+  ['AUCTION_BID_CREATED', [AuctionBidCreated]],
+])
+
+export default async function notification(
+  req: NextApiRequest,
+  res: NextApiResponse,
+): Promise<void> {
+  const liteflowSecret = process.env.LITEFLOW_WEBHOOK_SECRET
+  invariant(liteflowSecret, 'LITEFLOW_WEBHOOK_SECRET is required')
```
NicolasMahe, line 30:
```suggestion
  invariant(liteflowSecret, 'env LITEFLOW_WEBHOOK_SECRET is required')
```

NicolasMahe, line 30: Those two lines could be put outside the function for clarity. It doesn't seems to throw this error with the `build` or `dev` command, only when requesting this endpoint.

NicolasMahe (approved):

pages/api/notification.ts
antho1404: I think having it optional could make sense but ok to force it. In general I would like to also stop to rely on the global state for these kind of check but right now we have it this way so let's continue like that

pages/api/notification.ts
antho1404: moved as global check

package.json
```diff
@@ -26,6 +26,7 @@
     "@nft/chat": "^0.1.8",
     "@nft/email-connector": "^1.0.0-beta.11",
     "@nft/hooks": "^1.0.0-beta.11",
+    "@nft/webhook": "^1.0.0-beta.11-prerelease-2",
```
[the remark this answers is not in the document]
antho1404: Let's not block this PR, we are merging to dev, for sure we will need to update the library but that needs to be done before we release a new version

pages/api/notification.ts
```diff
@@ -0,0 +1,57 @@
...
+
+const emails = new Map<
+  keyof Events,
+  ((data: Events[keyof Events]) => SendMailOptions | null)[]
+>([
+  ['BID_CREATED', [BidCreated]],
+  ['AUCTION_BID_CREATED', [AuctionBidCreated]],
+])
+
+export default async function notification(
```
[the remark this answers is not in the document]
antho1404: I think the current naming is fine, if we change like that mean we need to change the file name to be coherent meaning the URL will be `/api/emailNotification` which is weird

PR #140, opened 2023-01-23
Migrate node env to lts

antho1404:
Update github actions to use the latest LTS version of node v18

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
       - name: Install Node
         uses: actions/setup-node@v2
         with:
-          node-version: 16
+          node-version: 18
           cache: 'npm'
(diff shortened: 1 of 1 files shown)


NicolasMahe:
I remember having issue with Vercel on a new project because it sets Node 18 by default. Can you make sure to deploy is working on vercel with node 18?

antho1404:
Previews are working fine, and v18 is now the lts so I don't think we should have any errors. Also, this is just for the github action, not vercel. When pushing that on vercel then yes we will have an issue until we change the node version in the settings of the project.

I pushed this change in https://github.com/liteflow-labs/starter-kit/pull/139, and the build is "working" for the "demo-bsc-testnet" that I changed to v18 (still have other build issue unrelated)

NicolasMahe (approved):

Issue #92, opened 2022-12-14
Implement pagination on ownerships


NicolasMahe:
Need to implement a pagination when fetching owners of an asset on at least the page token detail.

The query in database takes around 2s to 3s to execute.

https://github.com/liteflow-labs/starter-kit/blob/aafa50676ea0a989a4db47042203c37e7108ef73/pages/tokens/%5Bid%5D/index.gql#L45-L51

Check for similar query in the codebase.

EmmanuelDrouin:
![](https://github.trello.services/images/mini-trello-icon.png) [Implement pagination on ownerships](https://trello.com/c/fspjYwQg/515-implement-pagination-on-ownerships)

NicolasMahe:
done in https://github.com/liteflow-labs/starter-kit/pull/148

PR #148, opened 2023-01-27
Paginate results for owners

antho1404:
### Project organization

- Closes https://github.com/liteflow-labs/starter-kit/issues/92

### Description

Limit the size of queries by only fetching required owners and not all the owners for a specific asset.
All owners are loaded when needed and paginated when opening the dedicated modal about ownership

### DIFF
--- a/components/Pagination/Pagination.tsx
+++ b/components/Pagination/Pagination.tsx
@@ -4,13 +4,23 @@ import { IoChevronForward } from '@react-icons/all-files/io5/IoChevronForward'
 import { useMemo } from 'react'
 import Select from '../Select/Select'
 
-export type IProp = {
+type PropWithSelector = {
+  hideSelectors?: false | undefined
   limits: number[]
+  onLimitChange: (limit: string) => void
+}
+
+type PropWithoutSelector = {
+  hideSelectors: true
+  limits?: never
+  onLimitChange?: never
+}
+
+export type IProp = {
   limit: number
   page: number
   total?: number
   onPageChange: (page: number) => void
-  onLimitChange: (limit: string) => void
   result: {
     label: string
     caption: (context: {
@@ -20,18 +30,14 @@ export type IProp = {
     }) => string | JSX.Element
     pages: (context: { total: number }) => string | JSX.Element
   }
-  simple?: boolean
-}
+} & (PropWithSelector | PropWithoutSelector)
 
 export default function Pagination({
-  onPageChange,
-  onLimitChange,
-  total,
-  page = 1,
-  limits,
   limit,
+  onPageChange,
+  page,
   result,
-  simple,
+  total,
   ...props
 }: IProp): JSX.Element {
   const goTo = (newPage: number) => {
@@ -51,7 +57,7 @@ export default function Pagination({
   if (!total) return <></>
   return (
     <Flex
-      direction={{ base: simple ? 'row' : 'column', md: 'row' }}
+      direction={{ base: props.hideSelectors ? 'row' : 'column', md: 'row' }}
       align="center"
       justify={{ base: 'center', sm: 'space-between' }}
       w="full"
@@ -64,13 +70,13 @@ export default function Pagination({
         w={{ base: 'full', sm: 'auto' }}
         direction={{ base: 'column', sm: 'row' }}
       >
-        {!simple && (
+        {!props.hideSelectors && (
           <Select
             selectWidth={24}
             label={result.label}
             name="limit"
-            onChange={(e: any) => onLimitChange(e)}
-            choices={limits.map((x) => ({
+            onChange={(e: any) => props.onLimitChange(e)}
+            choices={props.limits.map((x) => ({
               value: x.toString(),
               label: x.toString(),
             }))}
@@ -99,7 +105,7 @@ export default function Pagination({
         gap={6}
         aria-label="Pagination"
       >
-        {!simple && (
+        {!props.hideSelectors && (
           <Flex align="center" gap={3}>
             <Select
               selectWidth={24}
--- a/components/Pagination/Pagination.tsx
+++ b/components/Pagination/Pagination.tsx
@@ -20,6 +20,7 @@ export type IProp = {
     }) => string | JSX.Element
     pages: (context: { total: number }) => string | JSX.Element
   }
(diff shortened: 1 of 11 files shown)


EmmanuelDrouin:
[High priority - Implement pagination on ownerships](https://trello.com/c/fspjYwQg/515-high-priority-implement-pagination-on-ownerships)

pages/index.tsx
```diff
@@ -163,6 +163,10 @@ const HomePage: NextPage<Props> = ({
           sales={asset.sales.nodes.map(convertSaleFull)}
           creator={convertUser(asset.creator, asset.creator.address)}
           owners={asset.ownerships.nodes.map(convertOwnership)}
+          ownerCount={parseInt(
+            asset.ownerships.aggregates?.sum?.quantity || '0',
```
NicolasMahe, line 167: This is not the owner count but the total number of asset owned by anybody.
It should use `ownerships.totalCount`.

components/Token/Header.tsx
```diff
@@ -32,6 +32,7 @@ export type Props = {
   currencies: SaleDetailProps['currencies']
   creator: TokenAssetProps['creator']
   owners: TokenAssetProps['owners']
+  ownerCount: TokenAssetProps['ownerCount']
```
NicolasMahe, line 35: Have "owner" as singular bring some confusion.
Is it the number of this asset owned by this owner? Or is it the total number of owners for this asset?
Could name it to something like:
```suggestion
  ownersCount: TokenAssetProps['ownersCount']
```
Or even
```suggestion
  numberOfOwners: TokenAssetProps['numberOfOwners']
```
This comment apply everywhere `ownerCount` is used.

pages/tokens/[id]/index.tsx
```diff
@@ -430,8 +430,13 @@ const DetailPage: NextPage<Props> = ({
           </Flex>
 
           <TokenMetadata
+            assetId={asset.id}
             creator={creator}
             owners={owners}
+            ownerCount={parseInt(
+              asset.ownerships.aggregates?.sum?.quantity || '0',
```
Comment (NicolasMahe) [inline pages/tokens/[id]/index.tsx:437]: This is not the owner count but the total number of asset owned by anybody.
It should use `ownerships.totalCount`.

components/Token/Header.tsx
antho1404: renamed everywhere to `numberOfOwners`

pages/tokens/[id]/index.tsx
Comment (antho1404) [reply pages/tokens/[id]/index.tsx]: fixed

NicolasMahe:
If you want to test with real data, set this env:
```
NEXT_PUBLIC_GRAPHQL_URL="https://api.blackmarket.defydisrupt.io/graphql"
NEXT_PUBLIC_CHAIN_ID="137"
NEXT_PUBLIC_NETWORK_NAME="Polygon"
```
and those domain for the next image:
```
'defylabs.mypinata.cloud',
'cdn.prod.defylabs.xyz',
's3.ap-southeast-2.amazonaws.com',
```
and then go to http://localhost:3000/tokens/137-0x86aad261465a1f7432efb8618d6736e910025c69-12

components/Token/Owners/Modal.tsx
```diff
@@ -55,20 +85,61 @@ const OwnersModal: VFC<Props> = ({ owners }) => {
...
+                    .map((owner) => (
+                      <OwnersModalItem key={owner.address} {...owner} />
+                    ))}
             </List>
           </ModalBody>
-          <ModalFooter as="div" />
+          <ModalFooter>
+            <ChakraPagination
+              pt="4"
+              limit={environment.PAGINATION_LIMIT}
```
NicolasMahe, line 119: We should set a pagination of 8 for this modal as it can displays max 8 elements without scroll:
Screenshot 2023-01-30 at 21 59 39
As the scroll bar is hidden and the last row is not truncated, it's actually hard to guess this area is scrollable.

components/Pagination/Pagination.tsx
```diff
@@ -20,6 +20,7 @@ export type IProp = {
     }) => string | JSX.Element
     pages: (context: { total: number }) => string | JSX.Element
   }
+  simple?: boolean
```
NicolasMahe, line 23: Not blocking but:
What this simple mean?
For what I can see by looking at the code, it actually hide the limit selector and the page selector.
Why not name it: `hideSelectors` or something more explicit?

Also, the props `limits`, `onLimitChange` and maybe `limit` could also be optional to remove the need to implement it in `components/Token/Owners/Modal.tsx` and simplify a bit of code.

NicolasMahe (approved):
Working good and solves the issue 🎊 
One easy improvement we could do on the number of items displayed.

components/Token/Owners/Modal.tsx
```diff
@@ -55,20 +85,61 @@ const OwnersModal: VFC<Props> = ({ owners }) => {
...
-              ))}
+              {loading
+                ? new Array(limit)
+                    .fill(0)
+                    .map((_, index) => (
+                      <ListItem
+                        key={index}
+                        image={<SkeletonCircle />}
+                        label={<SkeletonText noOfLines={2} />}
+                      />
```
NicolasMahe, line 107: I don't know if this can be improved but the height of the skeleton is not the same as the "real" rows. It changes the height of the modal everytime it loads for elements.

components/Token/Owners/Modal.tsx
antho1404: done [a4f771a](https://github.com/liteflow-labs/starter-kit/pull/148/commits/a4f771a0f852cd808d5a159d24fac36113131475)

components/Pagination/Pagination.tsx
antho1404: I like the `hideSelectors`. I had to come up with a kind of typescript hack that is not properly detected when using the components, but that should be fine (I think that's actually fixed in the new version of typescript with the satisfies and all)

components/Token/Owners/Modal.tsx
antho1404: The issue was that we didn't have the total number when fetching the next page, thus the pagination disappear and then the modal resizes. Fix that using the previous results total pages and also improve the width of the skeletons

Issue #155, opened 2023-02-02
New dynamic filters: explore + collections > UI update


EmmanuelDrouin:
Here is the summary of the work needed to finalize the UI update of the new dynamic filters for the explore + collection pages

### UI
- [x] Filters container
  - [x] Add max-width to prevent breaking the grid
    - Currently, the filter container resizes depending on the content inside
    - <img width="554" alt="Screen Shot 2566-02-02 at 14 41 10" src="https://user-images.githubusercontent.com/15095289/216261650-8b28f31d-02a6-4b29-85b3-ba6389fabc2c.png">
- [x] Radio button (status section)
  - [x] Add padding top-bottom > 4px
    - This will give the same height (40px) to the radio button as the other components (buttons, inputs, etc.)
  - [x] Radio active state > add missing background color brand/500-5
  - [x] Fix responsive > If the content of the radio button is bigger then the filters container, then the buttons that cannot fit the container should be display bellow in a new row
- [x] Currency selector (Price range) 
  - [x] There is a bug with the "currency selector", on click it opens the dropdown menu but pushes the selector outside of the container
  - https://user-images.githubusercontent.com/15095289/215677284-0fec8436-4767-4a50-b4c9-4ae33334ecce.mov
- [x] Collections
  - [x] Search field > update placeholder content for "Search by collections" (collections in plural)
  - [x] Collections list 
    - [x] Add hover effect (while hovering a collection)
      - Hover effect: Brand/500-5 color in background with transition ease
    - [x] Add missing padding left/right each element
    - [x] Fix layout with max-width and truncated state to prevent breaks (like the screenshot bellow)
      - <img width="523" alt="Screen Shot 2566-02-02 at 14 45 00" src="https://user-images.githubusercontent.com/15095289/216262382-4863fdf7-1c69-45ee-90d6-eda433122de9.png">
- [x] Properties
  - [x] Implement UI design for checkboxes properties 

### Responsive
- [x] Overall layout and container improvements to improve responsive of the component

Once on mobile breakpoint
- [x] By default filters are closed
- [x] On click on the filters button, open a full-modal (https://chakra-ui.com/docs/components/modal#modal-sizes) and display filters inside

EmmanuelDrouin:
![](https://github.trello.services/images/mini-trello-icon.png) [Collection work - Update filters component](https://trello.com/c/OhMsClLf/429-collection-work-update-filters-component)

antho1404:
[...]

PR #158, opened 2023-02-03
New dynamic filters for explore and collections

antho1404:
Closes https://github.com/liteflow-labs/starter-kit/issues/155

Redesign of the filter for the explore page and the collection page.
This filter includes the filter of type of offer, price, collections, and collection traits.
Update design for the filters and add modal for mobile


EmmanuelDrouin:
[Collection work - Update filters component](https://trello.com/c/OhMsClLf/429-collection-work-update-filters-component)

ismailToyran (approved):
Looking and working great.
There is only 1 issue I've faced.
Can you reset pagination to first page after using filters. Pagination is broken if you are not on first page and if there is only 1 page of items after filtering.

antho1404:
Merging this PR, I will fix the pagination in another PR https://github.com/liteflow-labs/starter-kit/issues/168 as I already identified the issue and it's an issue that if done correctly will have quite a lot of changes (simple tho) so better to have in a dedicated PR

Issue #162, opened 2023-02-06
Issue with gif and unlockable content


antho1404:
If a create an unlockable content with a GIF:
- I upload a GIF as the content to Unlock
- I can create my NFT without issue
- After the creation, the media crashes 
  - <img width="970" alt="Screen Shot 2566-02-06 at 12 09 35" src="https://user-images.githubusercontent.com/15095289/216889153-dd0a7028-769b-408f-87f8-271ebf936bcb.png">
  - https://nft-test-bsc-testnet-q6pizabgf-liteflow.vercel.app/tokens/97-0xcfbb45ff528c8ff6542887f631a1f704b92cf7db-60249402084937876423066029128237587855293854847399148942248025622398633406578
  - https://nft-test-bsc-testnet-q6pizabgf-liteflow.vercel.app/tokens/97-0xcfbb45ff528c8ff6542887f631a1f704b92cf7db-60249402084937876423066029128237587855293854847399148942288732701670928521081
- On a token detail page I can't interact with the section bellow the media to toggle the preview or hover the tooltip
  - <img width="624" alt="Screen Shot 2566-02-06 at 12 20 02" src="https://user-images.githubusercontent.com/15095289/216889750-26c41943-e465-4c64-b01f-4738c65fcd7b.png">

If I sign out of the platform, then the NFTs that have issue don't have any anymore and are properly displayed (the issue happen as an owner once connected)

**Connected**
<img width="1288" alt="Screen Shot 2566-02-06 at 12 21 12" src="https://user-images.githubusercontent.com/15095289/216889941-ba5f2057-888b-40d1-a858-4dbbcf2822e8.png">

**Not connected**
<img width="1288" alt="Screen Shot 2566-02-06 at 12 21 22" src="https://user-images.githubusercontent.com/15095289/216890129-4625eae0-9bb2-4b6a-81e7-4c60dd0c34dd.png">

_Originally posted by @EmmanuelDrouin in https://github.com/liteflow-labs/starter-kit/pull/161#pullrequestreview-1284564699_

antho1404:
@EmmanuelDrouin Could you upload the assets you used for that to reproduce

EmmanuelDrouin:
**Assets for** > https://nft-test-bsc-testnet-q6pizabgf-liteflow.vercel.app/tokens/97-0xcfbb45ff528c8ff6542887f631a1f704b92cf7db-60249402084937876423066029128237587855293854847399148942248025622398633406578

Preview
[...]

PR #164, opened 2023-02-07
Fix unlockable GIF

NicolasMahe:
### Project organization

- Closes https://github.com/liteflow-labs/starter-kit/issues/162

### Description

Fix display of unlockable GIF by bypassing the next/image optimisation rendering.

@EmmanuelDrouin I ask your review to double-check this PR fix the display of the gif with your assets.

### How to test

- Create an asset with any gif as unlockable content
- Check the gif is loaded properly
- Check the gif is not loaded properly with the dev branch

### Checklist

### DIFF
--- a/components/Image/Image.tsx
+++ b/components/Image/Image.tsx
@@ -12,7 +12,8 @@ const Image = chakra(NextImage, {
       'quality',
       'placeholder',
       'blurDataURL',
-      'loader ',
+      'loader',
+      'unoptimized',
     ].includes(prop),
 })
 
--- a/components/Token/Media.tsx
+++ b/components/Token/Media.tsx
@@ -19,6 +19,7 @@ const TokenMedia: VFC<{
   controls,
 }) => {
   const { colors } = useTheme()
+
   // prioritize unlockedContent
   if (unlockedContent) {
     if (unlockedContent.mimetype?.startsWith('video/'))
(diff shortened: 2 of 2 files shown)


antho1404 (approved):
Just adding a bit more context, there is a discussion about that on next. https://github.com/vercel/next.js/discussions/20138#discussioncomment-3795922

This only create a vercel and not local.

It seems like an issue related to the cache of vercel, and one user has the same error with arweave that is quite similar to ipfs, one way to solve that differently could be to invalidate the cache by adding a random query param.

NicolasMahe:
> Just adding a bit more context, there is a discussion about that on next. [vercel/next.js#20138 (reply in thread)](https://github.com/vercel/next.js/discussions/20138#discussioncomment-3795922)
> 
> This only create a vercel and not local.
> 
> It seems like an issue related to the cache of vercel, and one user has the same error with arweave that is quite similar to ipfs, one way to solve that differently could be to invalidate the cache by adding a random query param.

I tried to add something in the url to make sure the cache is not hit, but same result.
On the dev branch, the response contains the following header `x-vercel-cache: MISS` as well as the error `INVALID_IMAGE_OPTIMIZE_REQUEST`.
I don't think it's the cache as the issue should also be prevent on normal image.
I'm really not sure what's the reason.

Issue #168, opened 2023-02-08
Fix pagination on filter change


antho1404:
Can you reset pagination to first page after using filters. Pagination is broken if you are not on first page and if there is only 1 page of items after filtering.

![image](https://user-images.githubusercontent.com/33982052/217270610-155a7798-442d-460f-9867-fc17459e214d.png)

_Originally posted by @ismailToyran in https://github.com/liteflow-labs/starter-kit/pull/158#pullrequestreview-1287201091_

NicolasMahe:
Fixed by https://github.com/liteflow-labs/starter-kit/pull/169

PR #169, opened 2023-02-08
Fix issue with shallow navigation

antho1404:
### Project organization

- Closes https://github.com/liteflow-labs/starter-kit/issues/168
- Related to the new filter's UI

### Description

The filters introduce new navigation with the [`shallow` option](https://nextjs.org/docs/routing/shallow-routing) that basically changes the URL without having to have a full query on the server side. This is great as it is then faster for the user and has a lot less cloud function consumption.
That said, because of that, a lot of pages that were designed with SSR first approach do not work anymore, and for example, in #168, the pagination was not up to date, and was still getting the initial page from the SSR query.

I moved all the retrieving of pagination and sort (filter was already done) on the client side, this way, all issues are solved, and we can now correctly get the params from SSR or the shallow redirect.

**Note** the last commit, cce46ea88721beb76ed0387d7d3f0045e38c08c4, is not required and can be reverted if it's too much, it's just to ensure that all the codebase is similar and that we will not have an issue in the future

### How to test

The best way is to go on the explore, navigate to the next page, and select a filter. Previously this would not redirect to page 1 or no page, now it should

### DIFF
--- a/pages/collection/[chainId]/[id].tsx
+++ b/pages/collection/[chainId]/[id].tsx
@@ -60,7 +60,9 @@ import useAssetFilterFromQuery, {
 import useEagerConnect from '../../../hooks/useEagerConnect'
 import useExecuteOnAccountChange from '../../../hooks/useExecuteOnAccountChange'
 import useFilterState from '../../../hooks/useFilterState'
+import useOrderByQuery from '../../../hooks/useOrderByQuery'
 import usePaginate from '../../../hooks/usePaginate'
+import usePaginateQuery from '../../../hooks/usePaginateQuery'
 import LargeLayout from '../../../layouts/large'
 
 type Props = {
@@ -68,12 +70,6 @@ type Props = {
   collectionAddress: string
   currentAccount: string | null
   now: string
-  // Pagination
-  limit: number
-  page: number
-  offset: number
-  // OrderBy
-  orderBy: AssetsOrderBy
   // Currencies
   currencies: Pick<Currency, 'id' | 'image' | 'decimals'>[]
 }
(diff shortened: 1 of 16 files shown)


pages/collection/[chainId]/[id].tsx
```diff
@@ -251,7 +243,7 @@ const CollectionPage: FC<Props> = ({
           return { ...acc, [value]: query[value] }
         }, {}),
         ...otherFilters,
-        page: undefined,
+        page: 1,
```
Comment (antho1404) [inline pages/collection/[chainId]/[id].tsx:246]: Reset page to 1 instead of undefined to always have a value

pages/explore/collections.tsx
```diff
@@ -67,42 +64,31 @@ export const getServerSideProps = wrapServerSideProps<Props>(
 
     const { data, error } = await client.query<FetchExploreCollectionsQuery>({
       query: FetchExploreCollectionsDocument,
-      variables: { limit, offset, filter: queryFilter },
+      variables: { limit, offset, filter: queryFilter, orderBy },
```
antho1404, line 67: Bug fix here, the order was not used on the SSR so the results were not organized correctly from SSR

ismailToyran: Great catch 👍

ismailToyran (approved):
Looking and working great.
After testing a lot the only issue I've found is about searching.
The issue only appears if you do as follows
- Go on any page with search
- Filter as you like and use pagination (simply choose anything but first page)
- Type something on search (which will result in 1 page of items max) and pagination will not reset, resulting in being on the same pagination level again.

To resolve;
We should reset pagination after every search as well.

Up to you if you want to fix that in this PR or not 😉

antho1404:
Nice catch, I believe we had this issue for quite some time but that's now fixed [b792541](https://github.com/liteflow-labs/starter-kit/pull/169/commits/b7925413befe50b569a3957416e2a4bb8acff758)

PR #166, opened 2023-02-07
Use proper img component

NicolasMahe:
### Project organization

- Dependant on https://github.com/liteflow-labs/starter-kit/pull/160

### Description

Implement image size everywhere.
- Used the `layout="fixed"` with `width` and `height` whenever the image is display as a square with always the same size
- Implement media query in `sizes` prop otherwise
- Use the custom Image component everywhere

### How to test

- Navigate on the different page
- Check the image loaded are not full width, hopefully very close to the rendered size (most of the time it's close to 2 size of the rendered size because of pixel density)

### Checklist

- [x] Base branch of the PR is `dev`

PR #170, opened 2023-02-09
Optimize images sizes

ismailToyran:
Trying to fix image sizes using this PR https://github.com/liteflow-labs/starter-kit/pull/160
Also https://github.com/liteflow-labs/starter-kit/pull/166

- Used chakra breakpoints for sizes that belongs to grid.
- Added `100vw` and `(min-width: 80em) 1216px` for banners
- Used 
```ts
429px, // for mobile //
(min-width: 30em) 50vw,
(min-width: 48em) 33vw,
(min-width: 62em) 25vw,
(min-width: 80em) 290px",
```
- All of the cards using same grid system (collection, token, user card) (4-3-2-1 desktop to mobile), so hardcoding same sizes seems to be the best usage for me

This is a test PR, changes could be added to https://github.com/liteflow-labs/starter-kit/pull/160 as I didn't want to push anything to that. Tried to check and test all `Image` components on the repo

### DIFF
--- a/components/Collection/CollectionCard.tsx
+++ b/components/Collection/CollectionCard.tsx
@@ -40,6 +40,12 @@ const CollectionCard: FC<Props> = ({ collection }) => {
             alt={collection.name}
             layout="fill"
             objectFit="cover"
+            sizes="
+            429px,
+            (min-width: 30em) 50vw,
+            (min-width: 48em) 33vw,
+            (min-width: 62em) 25vw,
+            (min-width: 80em) 290px"
           />
         )}
         <Box
--- a/components/Collection/CollectionCard.tsx
+++ b/components/Collection/CollectionCard.tsx
@@ -58,7 +58,9 @@ const CollectionCard: FC<Props> = ({ collection }) => {
             <Image
               src={collection.image}
               alt={collection.name}
-              layout="fill"
+              // Twice the image size for better quality
+              width={104}
+              height={104}
               objectFit="cover"
             />
           )}
--- a/components/Collection/CollectionCard.tsx
+++ b/components/Collection/CollectionCard.tsx
@@ -45,7 +45,7 @@ const CollectionCard: FC<Props> = ({ collection }) => {
             (min-width: 30em) 50vw,
             (min-width: 48em) 33vw,
             (min-width: 62em) 25vw,
-            (min-width: 80em) 290px"
+            (min-width: 80em) 292px"
           />
         )}
         <Box
--- a/components/Collection/CollectionCard.tsx
+++ b/components/Collection/CollectionCard.tsx
@@ -41,7 +41,7 @@ const CollectionCard: FC<Props> = ({ collection }) => {
             layout="fill"
             objectFit="cover"
             sizes="
-            429px,
+            100vw,
             (min-width: 30em) 50vw,
             (min-width: 48em) 33vw,
             (min-width: 62em) 25vw,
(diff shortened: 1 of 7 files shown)


ismailToyran:
Read the comments on other PR, but wanted to use chakra breakpoints which also uses `min-width` to make it similar with the codebase.

components/Collection/CollectionCard.tsx
```diff
@@ -58,7 +64,9 @@ const CollectionCard: FC<Props> = ({ collection }) => {
             <Image
               src={collection.image}
               alt={collection.name}
-              layout="fill"
+              // Twice the image size for better quality
```
NicolasMahe, line 67: Why forcing to load a bigger image?
Next/Image is already taking into account the pixel density.

ismailToyran: It reduces the loaded image size a lot so that results in quality loss, it doesn't matter on images that are kinda small (currency, avatar etc.) But on posts and collection images you can see the quality change if you try with normal vs 2x of the image size. Maybe I'm wrong and that only happens on local though, but I've seen image changes quite a lot between the two dimensions.

antho1404: This is indeed super blurry on my side as well with the normal size. Indeed it is not required to double the size as next will do that if needed (aka in DPR 2). But here the issue is in DPR 1 where it is really pixelized, a lot more that is should (you can compare with another DPR 1 image)

| DPR 1 | DPR 2 | 
| --- | --- |
| Screenshot 2023-02-09 at 19 11 15 | Screenshot 2023-02-09 at 19 11 29 |

I think optimization for small images are not great

components/Collection/CollectionCard.tsx
```diff
@@ -40,6 +40,12 @@ const CollectionCard: FC<Props> = ({ collection }) => {
             alt={collection.name}
             layout="fill"
             objectFit="cover"
+            sizes="
+            100vw,
+            (min-width: 30em) 50vw,
+            (min-width: 48em) 33vw,
+            (min-width: 62em) 25vw,
+            (min-width: 80em) 290px"
```
antho1404, line 48: The max size is 292px

components/Token/Media.tsx
```diff
@@ -75,6 +75,12 @@ const TokenMedia: VFC<{
           layout="fill"
           objectFit={fill ? 'cover' : 'scale-down'}
           unoptimized={unlockedContent?.mimetype === 'image/gif'}
+          sizes="
```
antho1404, line 78: This one does not have the same size as others because of the filter column but it's ok to keep it like that

antho1404 (approved):

components/Collection/CollectionCard.tsx
antho1404: fixed

components/Collection/CollectionCard.tsx
NicolasMahe: That not the way to solve the quality issue.
We should set the quality to a value higher than the default 75:
https://nextjs.org/docs/api-reference/next/legacy/image#quality

NicolasMahe: @antho1404 
> This is indeed super blurry on my side as well with the normal size. Indeed it is not required to double the size as next will do that if needed (aka in DPR 2). But here the issue is in DPR 1 where it is really pixelized, a lot more that is should (you can compare with another DPR 1 image)

Of course it's going to look blurry if you force your browser DPR at 1 but your monitor is still at DPR 2. You actually cannot test without a monitor a DPR 1. On a monitor at DPR 1, everything is going to look blur, text included.

PR #174, opened 2023-02-09
Revert doubling the image size

NicolasMahe:
### Project organization

- Closes 
- Related https://github.com/liteflow-labs/starter-kit/pull/170/files#r1101170738
- Dependant on 

### Description

Revert doubling the image size in 3 components.

It's not up to us to try to improve what the next/image component image rendering process is doing.
Moreover, why changing it in only 3 places? This quality degradation is present everywhere.
I feel this is the start of a rabbit hole of optimising the quality of all images.
The goal was the opposite: reducing the image load speed by loading smaller, less quality, images.

I try to set the quality prop to 100 to see if it make a big difference. I can see only a slight quality increase. You can check it on the collection's square image with this preview: https://nft-test-polygon-mumbai-fihqdgrot-liteflow.vercel.app/collection/80001/0x7c68c3c59ceb245733a2fdeb47f5f7d6dbcc65b3
I don't think it's worth it to change this quality prop. Let's implement correctly the tool we have, but keep the default values as much as possible.

### Checklist

### DIFF
--- a/components/Collection/CollectionCard.tsx
+++ b/components/Collection/CollectionCard.tsx
@@ -64,9 +64,8 @@ const CollectionCard: FC<Props> = ({ collection }) => {
             <Image
               src={collection.image}
               alt={collection.name}
-              // Twice the image size for better quality
-              width={104}
-              height={104}
+              width={52}
+              height={52}
               objectFit="cover"
             />
           )}
--- a/components/Collection/CollectionHeader.tsx
+++ b/components/Collection/CollectionHeader.tsx
@@ -153,9 +153,8 @@ const CollectionHeader: FC<Props> = ({ collection, explorer, reportEmail }) => {
             <Image
               src={collection.image}
               alt={collection.name}
-              // Twice the image size for better quality
-              width={248}
-              height={248}
+              width={128}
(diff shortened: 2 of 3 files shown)


components/Collection/CollectionCard.tsx
```diff
@@ -64,10 +64,10 @@ const CollectionCard: FC<Props> = ({ collection }) => {
             <Image
               src={collection.image}
               alt={collection.name}
-              // Twice the image size for better quality
-              width={104}
-              height={104}
+              width={52}
+              height={52}
               objectFit="cover"
+              quality={100}
```
NicolasMahe, line 70: @ismailToyran does it look better on your computer?
I don't have a monitor with a DPR of 1, so it's really hard for me to see any difference.

NicolasMahe: I can see the increase of quality slightly improve the image.
But honestly, I would prefer to just leave the default.

antho1404 (approved):
This is not ideal but indeed let's stick to what we should do without hack and maybe solve that later by changing the next loader or something like that (maybe even just an update of next will solve this)

PR #180, opened 2023-02-13
wagmi wallet integration

antho1404:
### Description

Integrate wagmi as wallet provider instead of the previous web3 react.
Added better management of user's connection to only be connected on the platform only when the wallet is connected AND the wallet is logged in to the API thanks to the `useAccount` hook.

### How to test

Connect the wallet on the platform on the login page or other and check that every feature work as expected.
Test when you cleanup the cookies (you should be asked to sign again)
Test to disconnect your wallet
Test to logout on the platform
Test to be on the wrong network
...

### Known issues

~~When refreshing from some pages (like the explore page) while being connected, the `onConnect` from wagmi is not triggered making the login flow "stuck"
Still debugging this part~~ Done thanks to @ismailToyran


EmmanuelDrouin:
[Implement Wagmi](https://trello.com/c/zOsJhrwy/423-implement-wagmi)

hooks/useEagerConnect.ts
```diff
@@ -1,66 +1,23 @@
...
   useEffect(() => {
-    if (active) {
-      setTried(true)
-    }
-  }, [active])
+    if (ready) setHasBeenReady(true)
+  }, [ready, setHasBeenReady])
 
-  return tried
+  console.log({ ready: hasBeenReady || ready })
```
ismailToyran, line 21: Kept this `console.log` on purpose?

hooks/useEagerConnect.ts
```diff
@@ -1,66 +1,23 @@
...
-        if (isAuthorized) {
-          const address = await connectors.injected.getAccount()
-          if (address?.toLowerCase() === connectedAddress?.toLowerCase()) {
-            void activate(connectors.injected, undefined, true)
-            return
-          }
-        }
+  const ready = useMemo(() => {
+    if (isReconnecting) return false
+    if (!isConnected) return true
```
ismailToyran, line 12: Doesn't this one needs to be `false`?

components/Login/Form.tsx
```diff
@@ -1,39 +1,41 @@
...
 import WalletWalletConnect from '../Wallet/Connectors/WalletConnect'
 
 type Props = {
-  activate: (
-    connector: AbstractConnector,
-    onError?: (error: Error) => void,
-    throwErrors?: boolean,
-  ) => Promise<void>
-  networkName: string
+  onActivate?: () => void
```
antho1404, line 11: This now becomes a simple callback for "notification" and not to actually activate

components/Login/Form.tsx
```diff
@@ -1,39 +1,41 @@
...
-  )
 
   const hasStandardWallet =
     connectors.injected || connectors.coinbase || connectors.walletConnect
 
   return (
     <>
-      {connectors.email && <WalletEmail activate={activate} />}
+      {connectors.email && (
+        <Stack
```
antho1404, line 25: Kept a similar design as the buttons as this library automatically opens the popup from magic.

pages/_app.tsx
```diff
@@ -167,65 +158,50 @@ function AccountProvider(
...
             }
           : {},
-        cache: new InMemoryCache({
-          typePolicies: {
-            Account: {
-              keyFields: ['address'],
-            },
-          },
-        }).restore(props.cache),
+        cache: new InMemoryCache({}).restore(props.cache),
```
ismailToyran, line 199: Compared to Admin panel we are missing
```ts
cache: new InMemoryCache({
  typePolicies: {
    Account: {
      keyFields: ['address'],
    },
  },
}).restore(props.cache),
```
Any reason to include on admin and not this one?

ismailToyran (commented):
Everything works great.
I've seen you've added new library for email connection, tried that one as well and that works great as well. I'll approve when you provide me some context on the conversations I've created.

hooks/useEagerConnect.ts
antho1404: just removed

hooks/useEagerConnect.ts
antho1404: No, but indeed it's weird, here, if we are not connected (aka the wallet connected), then it means we tried we don't need to try more. This boolean here is either we tried or not. When we know we tried, it means it's ready (even if the result means that we are not connected)

pages/_app.tsx
antho1404: Good catch, that was an error.
Just reverted

hooks/useSigner.ts
```diff
@@ -1,18 +1,20 @@
...
+  const { isLoggedIn } = useAccount()
 
   return useMemo(() => {
-    if (!library) return
-    if (!account) return
-    return library.getSigner(account)
-  }, [library, account])
+    if (!isLoggedIn) return
+    if (!data) return
+    return (data as Signer & TypedDataSigner) || undefined
```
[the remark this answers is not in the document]
antho1404: Not really, this is to type with the `TypedDataSigner`, the signer from wagmi is already an implementation of `Signer` from `@ethersproject/abstract-signer` so all good here

package.json
```diff
@@ -17,22 +17,17 @@
     "@apollo/client": "^3.5.9",
     "@bugsnag/js": "^7.18.0",
     "@bugsnag/plugin-react": "^7.18.0",
-    "@chakra-ui/react": "^1.8.8",
+    "@chakra-ui/react": "^1.8.9",
```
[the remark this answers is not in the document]
ismailToyran: Great find 🎉

ismailToyran (approved):
Working as intended 👍 
Make sure to check my commits before merge though.

PR #181, opened 2023-02-13
Improve server side error management and redirection to not found errors

### LINKED ISSUE #111: Don't throw parse-query errors in service side functions
I think we should not throw error on data parse from URL / query.
Those errors are users' error, not app error.
On prod the user will receive a 500 errors with no explanation.

Example:
https://github.com/liteflow-labs/starter-kit/blob/7ce51793f07ea0f7a7e2edfaae810966566986d0/pages/users/%5Bid%5D/created.tsx#L51-L59

This errors should be 404 errors. Can use the simple `notFound` props:
```
return { notFound: true }
```
Example:
https://github.com/liteflow-labs/starter-kit/blob/aafa50676ea0a989a4db47042203c37e7108ef73/pages/tokens/%5Bid%5D/index.tsx#L124

ismailToyran:
### Project organization

- Closes #111 

### Description

- Return not found error on every request to a resource that is empty
- Remove the useless check on `data` (this can never be null/empty), probably from a previous version with wrong typing
- Did not throw any error for any list of items (aka list of assets, list of currency...) -- the app should handle that
- Did not throw any not found error for users, as even if they are not in the database, they can eventually be on the blockchain. Might need the API to always return (and validate) the address, and in that case, the not found will be good to add
- Add missing error management

### How to test

- Go to one of the pages and update the id in the URL, and you should have a 404
  - Checkout page
  - Token page
  - Collection page
  - Create form on a collection

### Checklist


ismailToyran:
I didn't change all `if (!data) throw new Error('data is falsy')` to `if (!data) return { notFound: true }` as we should get these values from the server. Just changed the ones that requires `assetId`, `id` etc.

pages/collection/[chainId]/[id].tsx
```diff
@@ -83,14 +82,14 @@ export const getServerSideProps = wrapServerSideProps<Props>(
...
-    invariant(chainIdStr, 'chainId is required')
+    if (!chainIdStr) return { notFound: true }
     const chainId = parseInt(chainIdStr, 10)
     const collectionAddress = ctx.params?.id
       ? Array.isArray(ctx.params.id)
         ? ctx.params.id[0]
         : ctx.params.id
       : null
-    invariant(collectionAddress, 'Collection Address is required')
+    if (!collectionAddress) return { notFound: true }
```
Comment (antho1404) [inline pages/collection/[chainId]/[id].tsx:92]: I left a comment on another PR where this is not entirely true, and in this case, we should throw an error as this means that next is not doing the routing properly.

This check should never occur as we don't validate the data we just retrieve it.

The flow should be:
- Retrieve the data from the params
- If they are supposed to be in the URL, and they don't exist, then throw an error (as next is broken somewhere)
- Process the data (doing a parseInt, for example)
- If there is an error, then ideally, we should throw a 400 (or ignore it and go to the next step)
- Check the presence of the resource (query the API)
- If the resource doesn't exist (because data mal formatted or the resource is inexistent), then throw a 404

That said, this might be a bit too much/overkill, and I'm ok to return some 404 in that case, even if that's not really correct, but I just wanted to highlight this.

If we could keep the 404 only at the last step, that would be great tho :)
Doing that will, I believe reduce drastically the size of this PR

pages/tokens/[id]/bid.tsx
```diff
@@ -103,6 +102,7 @@ export const getServerSideProps = wrapServerSideProps<Props>(
       },
     })
     if (chainCurrency.error) throw chainCurrency.error
+    if (!chainCurrency.data) return { notFound: true }
```
Comment (ismailToyran) [inline pages/tokens/[id]/bid.tsx:105]: This is also important data.
We should throw 404 if we don't get currency data right?

pages/collection/[chainId]/[id].tsx
Comment (ismailToyran) [reply pages/collection/[chainId]/[id].tsx]: Let me know if I get the idea correctly with the most recent commit please 😄

pages/collection/[chainId]/[id].tsx
```diff
@@ -175,11 +175,9 @@ export const getServerSideProps = wrapServerSideProps<Props>(
       })
 
     if (collectionDetailsError) throw collectionDetailsError
-    if (!collectionDetailsData)
-      throw new Error('collectionDetailsData is falsy')
-    if (!collectionDetailsData.collection) return { notFound: true }
+    if (!collectionDetailsData?.collection) return { notFound: true }
     if (collectionAssetsError) throw collectionAssetsError
-    if (!collectionAssetsData) throw new Error('collectionAssetsData is falsy')
+    if (!collectionAssetsData?.assets) return { notFound: true }
```
Comment (antho1404) [inline pages/collection/[chainId]/[id].tsx:180]: not sure this one is needed as the resource here is the collection, but i'm ok to keep it as anyway this should never happen, it is just a ts check

Comment (antho1404) [reply pages/collection/[chainId]/[id].tsx:180]: On second thoughts, let's revert this as it is a test on the result of the API `collectionAssetsData` that is and should always be present. Similar for the previous block of code. The initial file was correct, throwing an error when the API returns something not right (which should never happen) and then redirect to a 404 when the result of this request is empty

pages/create/[chainId]/[collectionAddress].tsx
```diff
@@ -81,7 +81,7 @@ export const getServerSideProps = wrapServerSideProps<Props>(
       },
     })
     if (error) throw error
-    if (!data) throw new Error('data is falsy')
+    if (!data) return { notFound: true }
```
Comment (antho1404) [inline pages/create/[chainId]/[collectionAddress].tsx:84]: This one should throw an error as this `data` should always be present if we don't have an error. This is a typescript check only.
We could add a not found on the collection tho with `if (!data.collection) return { notFound: true }` additionally to the existing check

antho1404 (commented):
I think there is a confusion here on the data retrieved from the api call and the actual resource we are looking for.
We need to have a 404 if the resource does not exists. If the api does not return any data for whatever reason we should definitely throw an error as this should never be the case.

Also I think we should have the following rule:
- If it's a single resource and this resource does not exists then we redirect to 404
- If it's a list of resources and this list is empty then we do nothing (list of assets, currencies...)

pages/create/[chainId]/[collectionAddress].tsx
Comment (ismailToyran) [reply pages/create/[chainId]/[collectionAddress].tsx]: We have `Empty` state for this one, so we shouldn't throw `404` for this one.
Just changed it to throw `new Error` if there is no data 👍

pages/account/wallet.tsx
```diff
@@ -22,6 +22,7 @@ export const getServerSideProps = wrapServerSideProps(
       query: WalletCurrenciesDocument,
     })
     if (error) throw error
+    if (!data) throw new Error('data is falsy')
     if (!data.currencies?.nodes) return { notFound: true }
```
antho1404, line 26: This one is not right, we could have no currency which would be useless but here it does not represent a resource but a list of resource so we shouldn't throw a not found

pages/account/wallet.tsx
```diff
@@ -22,6 +22,7 @@ export const getServerSideProps = wrapServerSideProps(
       query: WalletCurrenciesDocument,
     })
     if (error) throw error
+    if (!data) throw new Error('data is falsy')
```
antho1404, line 25: I don't think that's necessary, the type cannot be null nor undefined, so this test should be useless (if we trust typescript but let's do that)

pages/checkout/[id].tsx
```diff
@@ -71,6 +71,7 @@ export const getServerSideProps = wrapServerSideProps<Props>(
       },
     })
     if (error) throw error
+    if (!data) throw new Error('data is falsy')
```
Comment (antho1404) [inline pages/checkout/[id].tsx:74]: Here again, this cannot be null

pages/collection/[chainId]/[id].tsx
```diff
@@ -175,9 +175,11 @@ export const getServerSideProps = wrapServerSideProps<Props>(
       })
 
     if (collectionDetailsError) throw collectionDetailsError
+    if (!collectionDetailsData)
```
Comment (antho1404) [inline pages/collection/[chainId]/[id].tsx:178]: this cannot be null

pages/collection/[chainId]/[id].tsx
```diff
@@ -175,9 +175,11 @@ export const getServerSideProps = wrapServerSideProps<Props>(
...
+      throw new Error('collectionDetailsData is falsy')
     if (!collectionDetailsData?.collection) return { notFound: true }
     if (collectionAssetsError) throw collectionAssetsError
-    if (!collectionAssetsData?.assets) return { notFound: true }
+    if (!collectionAssetsData) throw new Error('collectionAssetsData is falsy')
```
Comment (antho1404) [inline pages/collection/[chainId]/[id].tsx:182]: This cannot be null

pages/create/[chainId]/[collectionAddress].tsx
```diff
@@ -82,6 +82,7 @@ export const getServerSideProps = wrapServerSideProps<Props>(
     })
     if (error) throw error
     if (!data) throw new Error('data is falsy')
```
Comment (antho1404) [inline pages/create/[chainId]/[collectionAddress].tsx:84]: data cannot be null, I guess we had all these checks before because of another client or something like that but can safely be removed everywhere if the type cannot be null or undefined

antho1404 (commented):
A few comments again, if you want I can take over this PR and propose a version to review, that might be easier to understand

ismailToyran:
> A few comments again, if you want I can take over this PR and propose a version to review, that might be easier to understand

Would be great indeed, if you can do that please 👍

antho1404 (approved):
@ismailToyran feel free to have a look and ask if you have any questions, I update the description to give more context.
Also feel free to merge if this is good for you

ismailToyran (commented):
Just one question, can merge otherwise 👍

antho1404:
closing in favor of https://github.com/liteflow-labs/starter-kit/pull/236 as this will be handled differently

PR #182, opened 2023-02-13
Display currencies depending on the chains

antho1404:
Dependency: https://github.com/liteflow-labs/starter-kit/pull/180

### Description

Ensure that the currencies display are the currencies for the current chain (aka the chain of the asset/collection) and not all the chain of the platform (in case of multi-chain)

### How to test

Import collections from 2 different chains and check that the currencies are different on asset from these 2 different chains

### DIFF
--- a/hooks/useChainCurrencies.ts
+++ b/hooks/useChainCurrencies.ts
@@ -7,7 +7,7 @@ import {
 
 export default function useChainCurrencies(
   chainId?: number,
-  onlyERC20 = false,
+  { onlyERC20 }: { onlyERC20?: boolean } = {},
 ): ReturnType<typeof useChainCurrenciesQuery> {
   const filter = {
     chainId: { equalTo: chainId } as IntFilter,
--- a/hooks/useChainCurrencies.ts
+++ b/hooks/useChainCurrencies.ts
@@ -1,0 +1,22 @@
+import {
+  AddressFilter,
+  CurrencyFilter,
+  IntFilter,
+  useChainCurrenciesQuery,
+} from '../graphql'
+
+export default function useChainCurrencies(
+  chainId?: number,
+  onlyERC20 = false,
+): ReturnType<typeof useChainCurrenciesQuery> {
+  const filter = {
+    chainId: { equalTo: chainId } as IntFilter,
+  } as CurrencyFilter
+
+  if (onlyERC20) filter['address'] = { isNull: false } as AddressFilter
+
+  return useChainCurrenciesQuery({
+    variables: { filter },
+    skip: chainId === undefined,
+  })
+}
--- a/pages/tokens/[id]/bid.tsx
+++ b/pages/tokens/[id]/bid.tsx
@@ -31,15 +31,22 @@ import {
 } from '../../../convert'
 import environment from '../../../environment'
 import {
+  AddressFilter,
   BidOnAssetDocument,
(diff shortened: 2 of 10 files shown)


pages/tokens/[id]/bid.tsx
```diff
@@ -120,6 +140,8 @@ const BidPage: NextPage<Props> = ({ now, assetId, meta, currentAccount }) => {
   })
   useExecuteOnAccountChange(refetch, ready)
 
+  const currencyRes = useChainCurrencies(data?.asset?.chainId, true)
```
Comment (ismailToyran) [inline pages/tokens/[id]/bid.tsx:143]: One suggestion to make it maybe more readable, not mandatory though. 
```suggestion
  const currencyRes = useChainCurrencies({ chainId: data?.asset?.chainId, onlyERC20: true })
```

pages/tokens/[id]/index.tsx
```diff
@@ -167,9 +187,16 @@ const DetailPage: NextPage<Props> = ({
       address: (ready ? address : currentAccount) || '',
     },
   })
+  const chainCurrency = useChainCurrencies(
+    data?.asset?.collection.chainId,
+    true,
+  )
```
Comment (ismailToyran) [inline pages/tokens/[id]/index.tsx:193]: Same as my other comment, not mandatory

ismailToyran (approved):
Looking good to me 👍 
Just one suggestion but not required to merge this one 🎉

pages/tokens/[id]/bid.tsx
Comment (antho1404) [reply pages/tokens/[id]/bid.tsx]: updated with `useChainCurrencies(chainId, { onlyERC20: true })` as `chainId` is required to be passed while `onlyERC20` is not

pages/tokens/[id]/bid.tsx
```diff
@@ -84,6 +91,19 @@ export const getServerSideProps = wrapServerSideProps<Props>(
...
+    >({
+      query: ChainCurrenciesDocument,
+      variables: {
+        filter: {
+          chainId: { equalTo: data.asset.chainId } as IntFilter,
+          address: { isNull: false } as AddressFilter,
+        } as CurrencyFilter,
+      },
+    })
+    if (chainCurrency.error) throw chainCurrency.error
```
Comment (NicolasMahe) [inline pages/tokens/[id]/bid.tsx:106]: If there is a custom hook `useChainCurrencies`, this query could also have it's function version to avoid possible error and confusion

hooks/useChainCurrencies.ts
```diff
@@ -0,0 +1,22 @@
...
+  const filter = {
+    chainId: { equalTo: chainId } as IntFilter,
+  } as CurrencyFilter
+
+  if (onlyERC20) filter['address'] = { isNull: false } as AddressFilter
+
+  return useChainCurrenciesQuery({
+    variables: { filter },
+    skip: chainId === undefined,
+  })
```
NicolasMahe, line 21: Could simplify returned data structure as this is a custom hook and not autogenerated Apollo hook:
```suggestion
  const {data} = useChainCurrenciesQuery({
    variables: { filter },
    skip: chainId === undefined,
  })
  return data?.currencies?.nodes
```
This will avoid to have this kind of useMemo:
```ts
  const currencies = useMemo(
    () => chainCurrency.data?.currencies?.nodes || [],
    [chainCurrency],
  )
```

NicolasMahe (approved):
Some code improvement to do but otherwise it's working good 👍

PR #185, opened 2023-02-14
Improve returned type of useChainCurrencies and add function to prefetch ChainCurrencies

NicolasMahe:
### Project organization

- Dependant on https://github.com/liteflow-labs/starter-kit/pull/182/files

### Description

- improve returned type of useChainCurrencies
- add function to prefetch ChainCurrencies

### DIFF
--- a/hooks/useChainCurrencies.ts
+++ b/hooks/useChainCurrencies.ts
@@ -1,22 +1,51 @@
+import { ApolloClient, NormalizedCacheObject } from '@apollo/client'
 import {
   AddressFilter,
+  ChainCurrenciesDocument,
+  ChainCurrenciesQuery,
+  ChainCurrenciesQueryVariables,
   CurrencyFilter,
   IntFilter,
   useChainCurrenciesQuery,
 } from '../graphql'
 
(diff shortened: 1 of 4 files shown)


hooks/useChainCurrencies.ts
```diff
@@ -1,22 +1,51 @@
...
+): [
+  NonNullable<ChainCurrenciesQuery['currencies']>['nodes'],
+  { error: Error | undefined },
+] {
+  const { data, error } = useChainCurrenciesQuery({
+    variables: { filter: filter(chainId, { onlyERC20 }) },
     skip: chainId === undefined,
   })
+
+  return [data?.currencies?.nodes || [], { error }]
```
antho1404, line 35: I don't think this is a good idea, ultimately, we should even have the loading state of this request to, for example, disable the select until all data are loaded. Maybe we would want to refetch etc...

antho1404:
I would revert the "improve returned type of useChainCurrencies" as this just adds limitations with little to no benefits. The "add function to prefetch ChainCurrencies" is good tho

Issue #90, opened 2022-12-14
Flash of Navigation when switching pages


achampagnedev:
When switching pages while logged in, the navigation state flashes from "logged out" state to "logged in" state 

Steps to reproduce: 
1. Login
2. Change pages
3. Notice state flash

https://user-images.githubusercontent.com/10590599/207502751-fbfb160f-d1bc-4406-b6b4-36d7e67f44db.mov

NicolasMahe:
I think this is a duplicate of https://github.com/liteflow-labs/marketplace-template/issues/77.
Please close this issue if it's the case.

NicolasMahe:
Let's keep it open then 👍

EmmanuelDrouin:
![](https://github.trello.services/images/mini-trello-icon.png) [Flash of Navigation when switching pages](https://trello.com/c/uhOUqaIC/516-flash-of-navigation-when-switching-pages)

NicolasMahe:
closed by https://github.com/liteflow-labs/starter-kit/pull/187

PR #187, opened 2023-02-14
Fix the flickering in the navbar when logged in

NicolasMahe:
### Project organization

- Closes https://github.com/liteflow-labs/starter-kit/issues/90

### Description

Fix the flickering in the navbar when logged in by also read the account from the previous data of the `useNavbarAccountQuery` hook

EDIT:
When implementing an alternative fix using a polling (https://github.com/liteflow-labs/starter-kit/pull/191), I find out that the NavBar query is executed when the page change even when removing the `refetch` and putting the `new Date(lastNotification || 0),` in memo to make sure the Date object is actually the same.
The `previousData` is working as expected, it make me think the issue is not related to the query or its input but somewhere else.
This is a very strange behavior.
Research moved to PR https://github.com/liteflow-labs/starter-kit/pull/192

### Checklist


antho1404:
This is, I believe, not the proper way to solve that, it does fix the issue, but it hides a more important issue of why this component is recreated. I believe the issue comes from the layout and most likely the cache when we set up the apollo client, which leads to recreating the whole client instead of updating it, leading to losing the cache from the navbar and thus having this flickering effect.

NicolasMahe:
I agree, it's not fixing the deep-down issue.

But if we look only at the logic of this component, because of the `refetch` called when router changes page, the `data` will return undefined during the refetch, thus the need to also use `previousData`.
I guess even if we fix the Apollo cache issue, we are going to keep the refetch logic, so this PR is still valid.

antho1404 (approved):
Let's merge this one, the problem is deeper than that but at least this does already improve things and is a good optimization anyway

PR #190, opened 2023-02-14
Remove useExecuteOnAccountChange hook

NicolasMahe:
### Description

Remove `useExecuteOnAccountChange` hook because all the places it's used pass the account to the query, so it will already refetch the query.


antho1404 (approved):
Perfect, I added a todo and I had this on my list of things I wanted to check. Indeed it was not necessary anymore :)

PR #192, opened 2023-02-14
Stop to execute Navbar query at every change of page

NicolasMahe:
### Project management

Based on https://github.com/liteflow-labs/starter-kit/pull/187

### Description

I'm trying to understand why the Navbar query is executed at every page change when the user is logged in.
This query is still executed at every page change even with the following modifications:
- remove refetch from navbar account query
- put lastNotification in a memo

@antho1404 Any idea where it's coming from?

### Checklist

### DIFF
--- a/components/Navbar/Navbar.tsx
+++ b/components/Navbar/Navbar.tsx
@@ -404,8 +404,8 @@ const Navbar: VFC<{
   const {
     data: accountData,
     refetch,
-    previousData: previousAccountData,
-  } = useNavbarAccountQuery({
+  const { data: accountData, previousData: previousAccountData } =
+    useNavbarAccountQuery({
     variables: {
       account: address?.toLowerCase() || '',
       lastNotification: new Date(lastNotification || 0),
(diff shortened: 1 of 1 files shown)


antho1404:
I have a good idea of where it's coming from, I answered on your other PR https://github.com/liteflow-labs/starter-kit/pull/187

NicolasMahe:
> I have a good idea of where it's coming from, I answered on your other PR #187. If I'm correct, you're on the right direction for the cache, but not entirely, this is mostly the client the issue and not the cache (even tho the cache can also be improved)

I did some investigation also to keep change the way the apollo client is initalised. Initialize it with no dep to `jwtToken` or `props.cache` to make sure it's not completely recreated. I then pass the `jwtToken` and the `props.cache` in 2 separate `useEffect`.
But it didn't improve the cache issue. The `restore` function from the cache is replacing all data in this cache:
> Restores the cache's state from a serialized NormalizedCacheObject (such as one returned by [extract](https://www.apollographql.com/docs/react/api/cache/InMemoryCache/#extract)). This removes all existing data from the cache.
> https://www.apollographql.com/docs/react/api/cache/InMemoryCache/#restore

I'm curious to see what you have in mind ;)

NicolasMahe:
closing this PR as it's old and prioritize

PR #193, opened 2023-02-15
Connection to any network based on the action

antho1404:
### Description

Currently, the application is required to be connected to a single network (and enforced by the layout). This PR adds a button wrapper for all actions network dependent on switching to the right network before an action lets the user connect with any chain and execute actions on many different chains.

### How to test

Switch your network to one not "supported" by one of the demo (meaning one network that does not have any collection), then try to execute the following actions:
- Create a listing
- Create a bid
- Accept a bid (from the detail page or account page)
- Cancel a bid (from the detail page or account page)
- Create an auction bid
- Accept an auction
- Create an nft

### DIFF
--- a/components/Button/SwitchNetwork.tsx
+++ b/components/Button/SwitchNetwork.tsx
@@ -1,0 +1,37 @@
+import { Button, ButtonProps } from '@chakra-ui/react'
+import { PropsWithChildren, useCallback } from 'react'
+import { useChainId, useSwitchNetwork } from 'wagmi'
+
+const ButtonWithNetworkSwitch = ({
+  children,
+  chainId,
+  ...props
+}: PropsWithChildren<
+  ButtonProps & {
+    chainId: number
+  }
+>): JSX.Element => {
+  const currentChainId = useChainId()
+  const { switchNetwork } = useSwitchNetwork()
+
+  const handleSwitchNetwork = useCallback(() => {
+    if (!switchNetwork) return
+    switchNetwork()
+  }, [switchNetwork])
+
+  if (currentChainId !== chainId)
+    return (
+      <Button
+        {...props}
+        onClick={handleSwitchNetwork}
+        type="button"
+        leftIcon={undefined}
+        rightIcon={undefined}
+      >
+        Switch Network
+      </Button>
+    )
+
+  return <Button {...props}>{children}</Button>
+}
+export default ButtonWithNetworkSwitch
--- a/components/Sales/Auction/Info.tsx
+++ b/components/Sales/Auction/Info.tsx
@@ -15,12 +15,14 @@ import { HiArrowNarrowRight } from '@react-icons/all-files/hi/HiArrowNarrowRight
 import useTranslation from 'next-translate/useTranslation'
 import { ReactElement, useCallback, useMemo, useState, VFC } from 'react'
 import { BlockExplorer } from '../../../hooks/useBlockExplorer'
+import ButtonWithNetworkSwitch from '../../Button/SwitchNetwork'
 import Link from '../../Link/Link'
 import AcceptAuctionModal from '../../Modal/AcceptAuction'
 import Price from '../../Price/Price'
 
 export type Props = {
   assetId: string
+  chainId: number
   auction: {
     id: string
     reserveAmount: BigNumber
@@ -44,6 +46,7 @@ export type Props = {
 const SaleAuctionInfo: VFC<Props> = ({
   signer,
   assetId,
+  chainId,
   auction,
   isOwner,
   isHomepage,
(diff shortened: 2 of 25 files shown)


components/Sales/Auction/Info.tsx
```diff
@@ -103,48 +106,63 @@ const SaleAuctionInfo: VFC<Props> = ({
...
-        action: {
-          children: (
+        action: (
+          <ButtonWithNetworkSwitch
+            chainId={chainId}
+            variant="outline"
+            colorScheme="gray"
+            bgColor="white"
+            onClick={acceptAuction}
+            isLoading={loading}
```
antho1404, line 158: Had to refactor this part and at the same time fix this loading state that was `loading` instead of `isLoading`

components/Button/SwitchNetwork.tsx
```diff
@@ -0,0 +1,37 @@
...
+}: PropsWithChildren<
+  ButtonProps & {
+    chainId: number
+  }
+>): JSX.Element => {
+  const currentChainId = useChainId()
+  const { switchNetwork } = useSwitchNetwork()
+
+  const handleSwitchNetwork = useCallback(() => {
+    if (!switchNetwork) return
```
NicolasMahe, line 18: > Some wallet apps do not support programmatic network switching and switchNetwork will be undefined. For those situations, you can typically switch networks in the wallet app and wagmi will stay up-to-date.
> https://wagmi.sh/react/hooks/useSwitchNetwork#usage

If we don't want to block user that use a wallet without this feature without any context, it should throw a toast with a message to tell the user to manually switch network.

components/Button/SwitchNetwork.tsx
```diff
@@ -0,0 +1,37 @@
...
+  if (currentChainId !== chainId)
+    return (
+      <Button
+        {...props}
+        onClick={handleSwitchNetwork}
+        type="button"
+        leftIcon={undefined}
+        rightIcon={undefined}
+      >
+        Switch Network
```
NicolasMahe, line 31: Could we show the network the user will be switched to?
Better UX and if there is an error, the user will be able to switch manually to the right network, now that it can read the network name.
```suggestion
        Switch to ${networks[chainId].name}
```

pages/_app.tsx
```diff
@@ -113,13 +109,6 @@ function Layout({
     ].filter(Boolean)
   }, [router.locale, userProfileLink])
 
-  // Automatically switch to the right network
-  useEffect(() => {
-    if (chainId === environment.CHAIN_ID) return
```
NicolasMahe, line 118: Can also remove env `NEXT_PUBLIC_CHAIN_ID` 🎊

components/Button/SwitchNetwork.tsx
antho1404: This starts to be really big for some action button

components/Button/SwitchNetwork.tsx
antho1404: Yes, that's a good idea to have a toast here that could solve your other comment as well, they click on the button and they have a toast to tell them to switch to XXX

pages/_app.tsx
antho1404: we do still need it for the "mintable collection" but this will most likely be changed soon to just use the `mintType` from the API

NicolasMahe (approved):
Pushed one fix https://github.com/liteflow-labs/starter-kit/pull/193/commits/359cd81c6b5f4f671b89dcd98de5df44a8b88f63 and some error improvement in the SwitchNetwork.
Working well for a first version 👍

components/Sales/Auction/Info.tsx
NicolasMahe: @ismailToyran I let you review this part please

components/Button/SwitchNetwork.tsx
```diff
@@ -0,0 +1,37 @@
+import { Button, ButtonProps } from '@chakra-ui/react'
```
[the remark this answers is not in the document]
antho1404: yes I saw that and was excited about doing that as well

components/Button/SwitchNetwork.tsx
antho1404: That's good, I just pushed an update to have the error handling directly in the hook and the chainId as well to have a more reactive approach

components/Button/SwitchNetwork.tsx
```diff
@@ -0,0 +1,49 @@
...
+        title: formatError(error),
+        status: 'error',
+      })
+    },
+  })
+  const toast = useToast()
+
+  const handleSwitchNetwork = useCallback(() => {
+    if (!switchNetwork) return
+    switchNetwork()
```
NicolasMahe, line 30: It doesn't work if you don't pass the chainId:
```suggestion
    switchNetwork(chainId)
```
I already fixed it in a previous commit. Any reason to revert it?

components/Button/SwitchNetwork.tsx
NicolasMahe: done in https://github.com/liteflow-labs/starter-kit/pull/194

components/Button/SwitchNetwork.tsx
antho1404: I added it to the `switchNetwork` params

ismailToyran (approved):
Implementation wise looks great but UX wise there might be something to improve later on.
If you are not logged in and on wrong network
- It shows `Place a bid`
- You click `Place a bid` and login modal appears
- You log in and `Place a bid` changes to `Switch Network`
- Now you need another modal popup to switch the network
- Then it changes back to `Place a bid` and you can actually start interacting.

To fix later on;
- We can chain these 2 actions maybe?
- Or we show switch network first, then show login modal?

Not blocking for this PR though.

components/Button/SwitchNetwork.tsx
NicolasMahe: Ok you moved it to a prop of `useSwitchNetwork` directly. It's working as expected

PR #194, opened 2023-02-15
Switch network with name

NicolasMahe:
### Project organization

- Dependant on https://github.com/liteflow-labs/starter-kit/pull/193

### Description

Show the name of the network to switch to in component SwitchNetwork

Screenshot 2023-02-15 at 15 29 33

### Checklist


antho1404:
This works ok in some elements but looks super bad in others. All the buttons in the accounts, the bids, modal, etc... are way too big, making everything super ugly, so I think the current solution is the best one for now. One thing that could be improved in terms of UX is that instead of having a long button with many information (maybe too much) then, when needed, display a modal with the instructions. That would be a much better solution than this

Screenshot 2023-02-15 at 23 05 56

NicolasMahe:
Indeed it's not looking good. I'm closing this PR. We can improve this logic later if we do an UI improvement of the login system

PR #196, opened 2023-02-15
Connect with the right network

NicolasMahe:
### Project organization

- Related https://github.com/liteflow-labs/starter-kit/pull/193#pullrequestreview-1299206560

### Description

Connect the wallet directly on the right network.
Just need to pass the `chainId` to the `useConnect` hook.

~~In draft because I want to try first if switching network is necessary for signature. If not, then the system should be more flexible and require this only when signing a transaction.~~ It does.

### How to test

- Disconnect your wallet
- Switch to a wrong network
- Try to place a bid
- Notice the change of network before connecting the account

### Checklist


antho1404 (approved):
Really nice :)

PR #197, opened 2023-02-15
Remove env NEXT_PUBLIC_NETWORK_NAME

NicolasMahe:
### Description

- Get network name from wagmi in Wallet page
- Remove NETWORK_NAME from env
- Display long wallet address in Wallet component

### How to test

- Go to page `account/wallet` while connected

### Checklist


antho1404:
I don't think this is good, this is even more misleading in the case of multichain support (with multi currencies), where we display all the currencies across all chains but change the warning based on the current network.

One simple way we could fix that now for the multi-chain is simply to remove this warning, as anything we would do with this warning will not fix the issue and might actually increase the misunderstanding like here.

@EmmanuelDrouin, what do you think as a quick fix to simply remove this warning and then we can work on a better way to display this wallet for a multi-chain environment

NicolasMahe:
> I don't think this is good, this is even more misleading in the case of multichain support (with multi currencies), where we display all the currencies across all chains but change the warning based on the current network.

The logic here is:as the wallet is connected to the displayed network, it should be safe for the user to transfer token to this wallet using this chain. But yeah it's not super logic with the multi-chain.

> One simple way we could fix that now for the multi-chain is simply to remove this warning, as anything we would do with this warning will not fix the issue and might actually increase the misunderstanding like here.
> 
> @EmmanuelDrouin, what do you think as a quick fix to simply remove this warning and then we can work on a better way to display this wallet for a multi-chain environment

I was thinking about this also but didn't want to change the UI. @EmmanuelDrouin please confirm.

EmmanuelDrouin:
> @EmmanuelDrouin, what do you think as a quick fix to simply remove this warning and then we can work on a better way to display this wallet for a multi-chain environment

@NicolasMahe @antho1404 Good to simply remove this warning for now as anyway the all page needs a re-design to be logical in a multi-chain context ;)

antho1404 (approved):
I removed the alert, this is now ready to merge

PR #199, opened 2023-02-16
Update useBlockExplorer to be multi chain compatible. Remove need for envs `NEXT_PUBLIC_BLOCKCHAIN_EXPLORER_URL` and `NEXT_PUBLIC_BLOCKCHAIN_EXPLORER_NAME`

antho1404:
### Description

Update the `useBlockExplorer` hook to accept a `chainId` and return the proper network based on this chain ID.

I had to create 2 new buttons `AcceptOffer` and `CancelOffer` as the modal was outside of an iteration, and I couldn't get the chain id to display the proper network.

### How to test

Create an offer, a bid, accept an offer, create an nft... and make sure that all the modals have the link to etherscan


NicolasMahe (approved):

components/Button/AcceptOffer.tsx
```diff
@@ -0,0 +1,71 @@
...
+    }
+  }, [accept, onClose, onOpen, onAccepted, offer, quantity, onError])
+
+  return (
+    <>
+      <ButtonWithNetworkSwitch
+        chainId={chainId}
+        {...props}
+        disabled={activeStep !== AcceptOfferStep.INITIAL}
+        isLoading={activeStep !== AcceptOfferStep.INITIAL}
```
ismailToyran, line 56: We shouldn't use `disabled`, instead use `isDisabled`
But in this case, as we already have `isLoading` we shouldn't even use `isDisabled`

components/Button/CancelOffer.tsx
```diff
@@ -0,0 +1,65 @@
...
+    }
+  }, [cancel, onClose, onOpen, onCanceled, offerId, onError])
+
+  return (
+    <>
+      <ButtonWithNetworkSwitch
+        chainId={chainId}
+        {...props}
+        disabled={activeStep !== CancelOfferStep.INITIAL}
+        isLoading={activeStep !== CancelOfferStep.INITIAL}
```
ismailToyran, line 50: We shouldn't use `disabled`, instead use `isDisabled`
But in this case, as we already have `isLoading` we shouldn't even use `isDisabled`

components/Button/AcceptOffer.tsx
NicolasMahe: @ismailToyran can you make the fix in a new PR please?

components/Button/CancelOffer.tsx
NicolasMahe: @ismailToyran can you make the fix in a new PR please?

components/Button/AcceptOffer.tsx
ismailToyran: https://github.com/liteflow-labs/starter-kit/pull/201

components/Button/CancelOffer.tsx
ismailToyran: https://github.com/liteflow-labs/starter-kit/pull/201

PR #201, opened 2023-02-16
Update usage of disable/isDisable

ismailToyran:
### Project organization
- Related to comments on https://github.com/liteflow-labs/starter-kit/pull/199#discussion_r1108144719

### Changes
- Replace `disabled` with `isDisabled`
- Remove disabled styles from `<Select />` component, as Chakra already applies those and instead use `isDisabled`
- Remove `disabled` props if it's same with `isLoading`, as both doing the same thing

### Description
As Chakra UI discourages to use `disabled` prop, instead we should use `isDisabled` prop.

### Checklist

### DIFF
--- a/components/Button/AcceptOffer.tsx
+++ b/components/Button/AcceptOffer.tsx
@@ -52,7 +52,6 @@ export default function AcceptOfferButton({
       <ButtonWithNetworkSwitch
         chainId={chainId}
         {...props}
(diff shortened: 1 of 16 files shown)


antho1404 (approved):

antho1404:
just pushed a tiny fix for formatting but otherwise good to merge

PR #204, opened 2023-02-20
Improve compatibility in token media component

NicolasMahe:
### Description

Improve compatibility in token media component.
- Add compatibility with image that are video
- Add compatibility with `blob` and `data` image
- Add support for base64 images

### Checklist


components/Token/Media.tsx
```diff
@@ -47,25 +93,22 @@ const TokenMedia: VFC<{
...
       />
     )
   }
-  if (image) {
-    if (imageError)
+
+  // display image
+  if (image && !imageError) {
+    // Use a basic image when the file is a blob or data
+    if (image.startsWith('blob:') || image.startsWith('data:'))
```
NicolasMahe, line 104: @antho1404 We didn't have the blob compatibility. How did it work with uploaded image?

antho1404: The blob was initially put in place to remove the optimization because next image didn't manage this type of image. For a couple of minor versions, they do support the blob, so that can and should go back to the default image, so we don't have yet another way to render this.

Also, the `data` should be supported by default, so we don't need another case again for that. I believe we have nothing to change here

ismailToyran:
- Changed `image: string | null | undefined` with `image: string` as image has to be present all the time
- Added `getUnlockedContentUrls` helper function

components/Token/Media.tsx
```diff
@@ -1,18 +1,51 @@
...
+  }
+  return {
+    image: imageUrl,
+    animation: animationUrl || null,
+  }
+}
+
 const TokenMedia: VFC<{
-  image: string | null | undefined
+  imageUrl: string
```
antho1404, line 40: This should be optional, we have nothing when we create a new nft in the `/create` page.

antho1404:
I don't see any improvement based on the info in the trello card with the following NFT that has an issue https://nft-test-uaan-liteflow.vercel.app/tokens/1-0x3b3ee1931dc30c1957379fac9aba94d1c48a5405-3

components/Token/Media.tsx
ismailToyran: Changed with [Add fallback bg for create page](https://github.com/liteflow-labs/starter-kit/pull/204/commits/4f265ad523a8f38013f4333f856f7cdec41b7569)
Next/Image expects an actual image, so passing `""` empty string is a bad usage.
Instead added `<Box bg="brand.50" />` for fallback to create page.

ismailToyran (commented):
[Fallback for mp4 import](https://github.com/liteflow-labs/starter-kit/pull/204/commits/cbdc787b5997dbf031bea7520ddccf88a943869f)
This was the actual issue.
I believe auto import adds mp4 or possible any other video to `imageUrl` so this component tries to render image instead of video. We should either add more checks or make some changes on backend so it will add these accordingly to correct fields while importing.

https://nft-test-uaan-9pipeehsj-liteflow.vercel.app/tokens/1-0x3b3ee1931dc30c1957379fac9aba94d1c48a5405-3
https://nft-test-uaan-9pipeehsj-liteflow.vercel.app/explore?search=Scarce+Resources

antho1404:
I reverted your fix, testing the extension is definitely not a proper way to do that, as file may not have extensions, also, what happened with other video files?...

The whole system had the entire fallback system that worked pretty well, the only issue we had was that the `onError` was not properly delegated to the next component and thus not triggering the fallbacks.

components/Token/Media.tsx
antho1404: This is not needed anymore by next but good to keep

antho1404 (approved):

PR #206, opened 2023-02-28
Redirection to metamask mobile

### LINKED ISSUE #51: Fix error when login with metamask on mobile
On mobile, to use metamask, users need to open the website from the metamask app.
Currently, we keep the metamask link, but it does nothing if open from a normal browser on mobile.
We should either redirect to the "download metamask" or have some ways to explain to the user that they need to open with the metamask app or simply disable/remove the button.

We can try to detect if metamask is present. If I remember correctly we can check the `window.ethereum` but it needs to be checked.

ismailToyran:
### Project organization
- Closes #51 
- Related to [trello issue](https://trello.com/c/xJUiC2YE/527-high-priority-fix-error-when-login-with-metamask-on-mobile)

### Description
Checks if `window.ethereum` is exists or not and depending on the return, if it's also mobile, it will now redirect user to metamask mobile app now.

### How to test
Visit the deployment link and try logging in on mobile device, it should redirect to playstore (tested on Android)

### Checklist
- [x] Base branch of the PR is dev
- [ ] Check on iphone


ismailToyran:
After some back and forth, I believe this is the correct way of implementing that.

components/Wallet/Connectors/Metamask.tsx
```diff
@@ -217,8 +221,38 @@ export const IconMetamask = (
...
+        <Text
+          as="span"
+          color="brand.black"
+          fontSize="sm"
+          fontWeight="semibold"
+          lineHeight={5}
+        >
+          Metamask
+        </Text>
+      </Box>
```
ismailToyran, line 252: Some duplication but didn't want to modify `<WalletBase />` component just for metamask

antho1404:
I cannot trigger it (on https://nft-test-bsc-testnet-2i05xzhh4-liteflow.vercel.app/), I'm on android, using chrome (as brave has a wallet integrated that bypasses metamask).
When I click on the button, nothing happens

antho1404:
Let's put in standby as this may be automatically solved when using rainbow kit

PR #207, opened 2023-02-28
Redirect to login page from notification page

### LINKED ISSUE #76: Redirect to login instead of throwing error if user is not connected on template Notification
We should redirect the user on the login page if user is not connected on `packages/templates/src/notification/Notification.tsx`. Error throwing is line 24

ismailToyran:
### Project organization
- Closes #76 
- Related to [trello card](https://trello.com/c/5EM3X6gH/520-high-priority-redirect-to-login-instead-of-throwing-error-if-user-is-not-connected-on-template-notification)

### Description
Redirects user to login page if they are not logged in and tries to navigate to notification page. Redirects back to notification after logging in

### How to test
Log out. Go to the url and add `/notification` It should redirect you to login page and after logging in back to notification page.
Also try logging out when on notification page, it should also redirect you to login page again.

### Checklist

### DIFF
--- a/pages/notification.tsx
+++ b/pages/notification.tsx
@@ -28,7 +28,13 @@ export const getServerSideProps = wrapServerSideProps<Props>(
   environment.GRAPHQL_URL,
   async (ctx, client) => {
     const address = ctx.user.address
-    if (!address) return { props: { currentAccount: null } }
+    if (!address)
+      return {
+        redirect: {
+          destination: '/login?redirectTo=notification',
+          permanent: false,
+        },
+      }
 
     const { data, error } = await client.query<GetNotificationsQuery>({
       query: GetNotificationsDocument,


antho1404:
I'm not sure what this PR fixes, I tested on the dev and the main, and we don't have any errors. 
The server does not throw any error, and the `useLoginRedirect` is doing its job of redirecting to the login. This improves the "flickering" part, but I just want to understand what this solves. If it's the flickering, then this should be done across the whole application, if it's the error, then this PR is not needed, and the issue can just be closed.

ismailToyran:
I agree, I think what I did is the correct usage of redirecting. And should be used across whole application.
I believe this issue was already fixed as this is an old issue, so I've tried to improve the usage.
Could be closed or we can keep this one open to implement this usage to other pages.

antho1404:
let's close this PR and the associated issue, hopefully with the SSR update this will not be needed anymore

PR #215, opened 2023-03-02
Normalize accountImage component usage

### LINKED ISSUE #87: Improve params of component AccountImage
The `AccountImage` image has an optional param `size` that default to `32`.
The codebase also define width and height of this composant with params `w` and `h`.

If they don't match, the component with a metamask generated icon is not displayed correctly.

Component:
https://github.com/liteflow-labs/nft/blob/27fcf7f1ea039038579faa2e9d2b5a5ec20a5cfe/packages/components/src/Wallet/Image.tsx#L8-L38

- In some places, only size is set, but not `w` and `h`:
https://github.com/liteflow-labs/nft/blob/79e7c9f152b4fa4f8b3d5d6537d0ca20f3d2f190/packages/components/src/Bid/Bid.tsx#L134-L140
https://github.com/liteflow-labs/nft/blob/d1f2d77c418a71edf1d723e43e0c4a46b2cf1e8c/packages/components/src/Navbar/Navbar.tsx#L356-L362

- In some other, only `w` and `h` with a value of 8 (that match the default size of 32px):
https://github.com/liteflow-labs/nft/blob/27fcf7f1ea039038579faa2e9d2b5a5ec20a5cfe/packages/components/src/Token/Owners/ModalActivator.tsx#L23-L27
https://github.com/liteflow-labs/nft/blob/27fcf7f1ea039038579faa2e9d2b5a5ec20a5cfe/packages/components/src/Token/Owners/ModalActivator.tsx#L35-L39
https://github.com/liteflow-labs/nft/blob/27fcf7f1ea039038579faa2e9d2b5a5ec20a5cfe/packages/components/src/Sales/Auction/Incomplete/FailedReservePrice.tsx#L51-L55
https://github.com/liteflow-labs/nft/blob/27fcf7f1ea039038579faa2e9d2b5a5ec20a5cfe/packages/components/src/Sales/Auction/Incomplete/Successfully.tsx#L55-L59
https://github.com/liteflow-labs/nft/blob/79e7c9f152b4fa4f8b3d5d6537d0ca20f3d2f190/packages/components/src/Sales/Direct/ModalItem.tsx#L78-L83

---

I would like to make sure the `size`, `h` and `w` are properly define.

Suggestions/questions to make it better
1. Make `size` params required and remove its default. This will force the developer to properly set the size
2. Is it necessary to put the `w` and `h` if the `size` is set?
  - If yes, could we force the `w` and `h` directly in the composant `AccountImage` based on the `size`? Careful, the `size` is in pixel but the `w` and `h` are in chakra unit (px/8). I would suggest to switch the `size` to chakra unit and multiply it by 8 when need pixel.

ismailToyran:
### Project organization
- Closes #87 
- Related [trello card](https://trello.com/c/z6jv0P5k/518-low-priority-improve-params-of-component-accountimage)

### Description
There has been changes on this component since this issue has been created.
But there were still different kind of usages that wasn't matching rest of the usage.
This PR normalizes the usage of `<AccountImage />` component across the application.

### How to test
Everything should be same, except navbar profile icon positioning and height. (Which wasn't centered before because of the wrong height)

### Checklist

### DIFF
--- a/components/Sales/Auction/Incomplete/FailedReservePrice.tsx
+++ b/components/Sales/Auction/Incomplete/FailedReservePrice.tsx
@@ -51,14 +51,11 @@ const SaleAuctionIncompleteReservePrice: VFC<Props> = ({
           {t('sales.auction.failed-no-reserve.highest-bid')}
         </Heading>
         <Flex align="center" gap={3}>
-          <Box
+          <Flex
             as={AccountImage}
-            display="inline-block"
-            h={8}
-            w={8}
-            rounded="full"
             address={bestBid.maker.address}
             image={bestBid.maker.image}
+            rounded="full"
           />
           <Heading as="h4" variant="heading2" color="brand.black">
             <Trans
--- a/components/Sales/Direct/ModalItem.tsx
+++ b/components/Sales/Direct/ModalItem.tsx
@@ -1,5 +1,4 @@
 import {
-  Box,
   Button,
   Flex,
   Icon,
@@ -85,16 +84,14 @@ const SaleDirectModalItem: VFC<Props> = ({
     <>
       <ListItem
         image={
-          <Flex as={Link} href={`/users/${sale.maker.address}`}>
-            <Box
+          <Link href={`/users/${sale.maker.address}`}>
+            <Flex
               as={AccountImage}
               address={sale.maker.address}
               image={sale.maker.image}
               size={40}
-              w={10}
-              h={10}
             />
-          </Flex>
+          </Link>
         }
         label={
           <Flex
--- a/components/Bid/Bid.tsx
+++ b/components/Bid/Bid.tsx
@@ -140,7 +140,7 @@ const Bid: VFC<Props> = ({
               address={bid.maker.address}
               image={bid.maker.image}
               size={40}
-              cursor="pointer"
+              rounded="full"
             />
           </Link>
         }
--- a/components/Navbar/Navbar.tsx
+++ b/components/Navbar/Navbar.tsx
@@ -339,15 +339,13 @@ const UserMenu: VFC<{
   return (
     <Menu>
       <MenuButton>
-        <Flex>
-          <Box
-            as={AccountImage}
-            rounded="full"
-            address={user.address || account}
-            image={user.image}
-            size={40}
-          />
-        </Flex>
+        <Flex
+          as={AccountImage}
+          address={user.address || account}
+          image={user.image}
+          size={40}
+          rounded="full"
+        />
       </MenuButton>
       <MenuList>
         <Link href={`/users/${account}`}>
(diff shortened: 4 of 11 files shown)


antho1404:
I guess now that the alignment is correct for the image, the notification button feels off, maybe it's not aligned as well.

Screenshot 2023-03-03 at 09 39 42

components/Sales/Direct/ModalItem.tsx
```diff
@@ -85,16 +84,14 @@ const SaleDirectModalItem: VFC<Props> = ({
     <>
       <ListItem
         image={
-          <Flex as={Link} href={`/users/${sale.maker.address}`}>
-            <Box
+          <Link href={`/users/${sale.maker.address}`}>
```
antho1404, line 87: My image is "cut" in this modal, where it is displayed correctly everywhere else. Maybe unrelated, happy to move that to a dedicated issue if needed.

Screenshot 2023-03-03 at 09 48 29

components/Sales/Auction/Incomplete/FailedReservePrice.tsx
```diff
@@ -51,14 +51,11 @@ const SaleAuctionIncompleteReservePrice: VFC<Props> = ({
           {t('sales.auction.failed-no-reserve.highest-bid')}
         </Heading>
         <Flex align="center" gap={3}>
-          <Box
+          <Flex
             as={AccountImage}
-            display="inline-block"
-            h={8}
-            w={8}
```
NicolasMahe, line 58: Let's not remove the w & h if you don't replace by size.
This will make the next/image to load the biggest image.

NicolasMahe: This comment apply to many other places

ismailToyran: `<AccountImage />` is using `defaultSize = 32` prop as default.
So in this case these are not needed at all.
Whole purpose of this PR is removing those values as they make no sense. If we use fixed width and height, it's wrapper shouldn't have those `w={8} h={8}` values. As it will be rendered `32px` image anyway (with next/image)

ismailToyran:
Also related to your comment
> I guess now that the alignment is correct for the image, the notification button feels off, maybe it's not aligned as well.



[Needs to be wrapped with flex because how next/image wraps it's content](https://github.com/liteflow-labs/starter-kit/pull/215/commits/5c1e590353646bea9b1774049c1b6392e90b274b)

antho1404 (approved):

components/Sales/Auction/Incomplete/FailedReservePrice.tsx
NicolasMahe: Perfect then :)

PR #218, opened 2023-03-06
Remove ownershipsExist and assetExists from filters

NicolasMahe:
### Project organization

- Closes 
- Related https://github.com/liteflow-labs/backend/issues/2406
- Dependant on 

### Description

Remove ownershipsExist and assetExists from filters to prepare for upcoming breaking change on the API.

### Checklist

### DIFF
--- a/pages/explore/explore.gql
+++ b/pages/explore/explore.gql
@@ -7,7 +7,7 @@ query FetchAllERC721And1155(
   $filter: [AssetFilter!]
 ) {
   assets(
-    filter: { ownershipsExist: true, and: $filter }
+    filter: { and: $filter }
     first: $limit
     offset: $offset
     orderBy: $orderBy
--- a/pages/collection/[chainId]/[id].gql
+++ b/pages/collection/[chainId]/[id].gql
@@ -44,7 +44,6 @@ query FetchCollectionAssets(
 ) {
   assets(
     filter: {
-      ownershipsExist: true
       chainId: { equalTo: $chainId }
       collectionAddress: { equalTo: $collectionAddress }
       and: $filter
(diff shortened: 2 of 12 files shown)


pages/explore/explore.gql
```diff
@@ -7,7 +7,7 @@ query FetchAllERC721And1155(
   $filter: [AssetFilter!]
 ) {
   assets(
-    filter: { ownershipsExist: true, and: $filter }
+    filter: { and: $filter }
```
antho1404, line 10:
```suggestion
    filter: $filter
```

NicolasMahe: Cannot pass an array of filter

antho1404: We should be able to change the gql param to `AssetFilter!` instead of `[AssetFilter!]`

NicolasMahe: I'm not sure to follow the logic. The function that creates the filters in `explore.tsx`, `convertFilterToAssetFilter`, is returning an array. Why do you want to change it?

NicolasMahe:
> Before merging and releasing, are all the "pending" assets already cleaned up in the different backends? If not this will display the "pending" assets

Done

pages/explore/explore.gql
antho1404: This filter can either be an object (where all the keys are used as and AND condition) or this array that we need to force the `and`. Having the object is just the way we use it everywhere else, so this would make the codebase consistent. Right now, we had this `and` just because we needed to enforce this `ownershipExists` all the time.

Definitely not a blocker but I would love to have some consistency in the codebase but happy to move that in an issue

antho1404: https://github.com/liteflow-labs/starter-kit/issues/219

PR #225, opened 2023-03-07
Notification page link redirection improvement

ismailToyran:
### Project organization
- Related [trello card](https://trello.com/c/beWElkHo/655-notification-link-redirection-improvement)

### Description
Adds new redirections for the items on notification page

### How to test
Test by clicking different type of notifications on the notification page. It should redirect user to the related page now instead of directing everything to the token detail page.

### Checklist

### DIFF
--- a/components/Notification/Detail.tsx
+++ b/components/Notification/Detail.tsx
@@ -27,6 +27,7 @@ import {
 
 export type IProps = {
   createdAt: Date
+  currentAccount: string | null
   action: NotificationAction
   accountVerification: {
     status: AccountVerificationStatus
@@ -47,7 +48,6 @@ export type IProps = {
     unitPrice: string
     quantity: string
     asset: {
-      id: string
       image: string
       name: string
     }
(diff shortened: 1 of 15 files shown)


antho1404 (commented):
I'm ok with this one if that's the direction. I just want to confirm with @EmmanuelDrouin as this is in my opinion not the best UX as we can now redirect to a page that does not contains the data we are looking for. If we click on a old notification about the bids we placed and the bid related to the notifications is 10 pages later, we will not see the bid of the notification.

EmmanuelDrouin:
> I'm ok with this one if that's the direction. I just want to confirm with @EmmanuelDrouin as this is in my opinion not the best UX as we can now redirect to a page that does not contains the data we are looking for. If we click on a old notification about the bids we placed and the bid related to the notifications is 10 pages later, we will not see the bid of the notification.

Yes I'm aware of this possible flow you are describing and it has been communicated to the client who requested these changes. I'm good to push further in the current direction as I still believe this improve the current UX

PR #229, opened 2023-03-08
Chat page added

ismailToyran:
### Project organization
- Related [trello card](https://trello.com/c/lYgYDlnW/656-messaging-system-starter-kit-responsive-improvement)

### Description
Removes chat widget from app level and add new chat page

### How to test
Either use navigation after logging in or go to the `/chat` url. Should work both for mobile and desktop.

### Checklist


antho1404:
Can we design-wise put this in a card or have some borders/shadow around, we have borders inside but not outside, which looks bad. Also, the height needs to be adjusted, right now, I cannot use it on my desktop, as when I open the page, I cannot see the messages and type a new one at the same time. The chat should be 100% of the container and not more otherwise it is not usable

ismailToyran:
https://github.com/liteflow-labs/chat/pull/31#issuecomment-1511011386 should be good enchancement for UI, as height is not lined up between left and right navbar

ismailToyran:
Ready for review once again. @antho1404 
Design has been approved by Emmanuel.
I've shown him what will change with https://github.com/liteflow-labs/chat/pull/31
So it will be great if we merge that and update package as well 👍

antho1404 (approved):
The box for the chat layout is not great but we cannot do much better, if we migrate with chakra 2.x we will be able to replace that with a nice card

PR #230, opened 2023-03-14
Use rainbow kit for login

antho1404:
### Description

Update the login system from a custom login page/modal to an already well-established tool https://www.rainbowkit.com/.
This will help in many ways:
- Better detection of the wallets present on the user's browser
- Better compatibility with wallets like nice connection to remote metamask, install of new wallets etc...
- Simplification of the codebase

### How to test

Go to any page and login with the "sign in" button in the nav bar or on the `/login` page

### Review
- @EmmanuelDrouin to confirm the ux flow
- @NicolasMahe review the code

### TODO
- [x] Fix issue that once connected wagmi `onConnect` function is not called https://github.com/wagmi-dev/wagmi/issues/1998


antho1404:
Note: The issue with the `onConnect` not called is related to the fact that `useAccount` is also using the `useWagmiAccount`. The custom `useAccount` should not use the wagmi's `useAccount`

components/Navbar/Navbar.tsx
```diff
@@ -524,12 +524,19 @@ const Navbar: VFC<{
                 signOutFn={() => logout().then(disconnect)}
               />
             </>
+          ) : isConnected ? (
+            <Button
+              colorScheme="brand"
+              isLoading
+              loadingText={t('navbar.signing-in')}
```
antho1404, line 531: Added a new state so while we are signing, the button is disabled and loading.

hooks/useAccount.ts
```diff
@@ -22,11 +21,9 @@ const COOKIE_OPTIONS = {
 }
 
 export default function useAccount(): AccountDetail {
-  const { address, isConnected } = useWagmiAccount()
```
antho1404, line 25: Fix related to the issue I mentioned.
This hook now only relies on the jwt in cookies and the actual state in the sdk.
When we load, if the `onConnect` in wagmi is not called then we never login to the api so we have the same behavior

environment.ts
```diff
@@ -92,6 +93,7 @@ const MINTABLE_COLLECTIONS = (
   .map((address) => ({ address: address.toLowerCase(), chainId: CHAIN_ID }))
 
 const environment: Environment = {
+  APP_NAME: 'Acme',
   MAGIC_API_KEY: process.env.NEXT_PUBLIC_MAGIC_API_KEY,
   PUBLIC_ETHEREUM_PROVIDER: process.env.NEXT_PUBLIC_ETHEREUM_PROVIDER,
```
NicolasMahe, line 98: This env should be removed because not used since Wagmi integration.
```suggestion
```

hooks/useEagerConnect.ts
```diff
@@ -3,8 +3,8 @@ import { useAccount as useWagmiAccount } from 'wagmi'
 import useAccount from './useAccount'
 
 export default function useEagerConnect(): boolean {
```
NicolasMahe, line 5: Is this hook still require? I don't see it on the admin.

components/Offer/Form/Bid.tsx
```diff
@@ -413,18 +409,13 @@ const OfferFormBid: FC<Props> = (props) => {
           </ButtonWithNetworkSwitch>
         </>
       ) : (
-        <Button size="lg" type="button" onClick={loginOnOpen}>
+        <Button size="lg" type="button" onClick={openConnectModal}>
```
NicolasMahe, line 412: It would be really nice to have the "Sign in" and "Signing in..." state on this button.
A new component similar to `ButtonWithNetworkSwitch` that manage the login, and optionally switch of network would be great! It could be use everywhere.

pages/login.tsx
```diff
@@ -67,16 +70,17 @@ const LoginPage: NextPage = () => {
...
         mt={12}
         mb={{ base: 12, lg: 24 }}
         justify="center"
       >
-        <LoginForm onActivate={handleAuthenticated} chainId={undefined} />
+        <ConnectButton
+          accountStatus="address"
+          chainStatus="none"
+          showBalance={false}
+        />
```
NicolasMahe, line 83: I guess we still need this page because of the referral system?

@EmmanuelDrouin Design is very simple but not bad: https://nft-test-uaan-git-feature-rainbowkit-integration-liteflow.vercel.app/login

hooks/useAccount.ts
```diff
@@ -70,7 +70,6 @@ export default function useAccount(): AccountDetail {
       address: jwt?.address?.toLowerCase(),
       jwtToken: jwt?.token,
       isLoggedIn: !!jwt,
-      isConnected: !!jwt,
```
NicolasMahe, line 73: If it's removed here let's remove it from the admin then: https://github.com/liteflow-labs/admin/blob/main/hooks/useAccount.ts#L87

connectors.ts
```diff
@@ -1,56 +1,95 @@
...
 
-export const connectors: {
-  injected?: InjectedConnector
-  walletConnect?: WalletConnectConnector
-  coinbase?: CoinbaseWalletConnector
-  email?: Connector
-} = {
-  email: environment.MAGIC_API_KEY
-    ? emailConnector(environment.CHAIN_ID)
-    : undefined,
```
NicolasMahe, line 31: The email connector was activated only if the `environment.MAGIC_API_KEY` is truly. Looks like you remove this behaviour.

hooks/useAccount.ts
NicolasMahe: Actually, why not keeping the `isConnected` that comes from `useWagmiAccount` instead of importing `useWagmiAccount` in the different components?
The goal is to centralize this kind of stuff in this hook.

environment.ts
```diff
@@ -92,6 +93,7 @@ const MINTABLE_COLLECTIONS = (
   .map((address) => ({ address: address.toLowerCase(), chainId: CHAIN_ID }))
 
 const environment: Environment = {
+  APP_NAME: 'Acme',
```
NicolasMahe, line 96: Is it really necessary to add a new config here?
There is already a bunch of "Acme" hardcoded in the code: header, footer, and referral page

NicolasMahe (commented):
Working pretty good 👍 Good job.
I'm a bit worry about the difference of implementation between admin and starter-kit. The goal is to make sure both are using the same hooks and same logic.

pages/login.tsx
EmmanuelDrouin: Simple but good for me 👍

EmmanuelDrouin (approved):

pages/login.tsx
antho1404: yes exactly, that's only referral related

environment.ts
antho1404: Reverted to keep some consistency, now it's hardcoded everywhere [fb81473](https://github.com/liteflow-labs/starter-kit/pull/230/commits/fb814730882316e62d30e95b4ca5e5f8d5638184)

hooks/useAccount.ts
antho1404: This issue is that if we keep the `useWagmiAccount`, the whole login does not work anymore because of a bug with wagmi I talked about here https://github.com/liteflow-labs/starter-kit/pull/230#issuecomment-1467868652.

When the bug is fixed, we will be able to add everything back, but until then, we need to do this workaround that I did here https://github.com/liteflow-labs/starter-kit/pull/230#discussion_r1139896373

NicolasMahe (approved):
I'm approving this PR because it's working but I would prefer to merge the `onConnect` fix from https://github.com/liteflow-labs/starter-kit/pull/233 first to limit the code diff between starter-kit and admin and track much easier the fix and its future removal.

hooks/useEagerConnect.ts
```diff
@@ -4,7 +4,7 @@ import useAccount from './useAccount'
...
   const { isReconnecting } = useWagmiAccount()
-  const { isConnected, isLoggedIn } = useAccount()
+  const { isLoggedIn, isConnected } = useAccount()
```
antho1404, line 7:
```suggestion
  const { isConnected, isLoggedIn } = useAccount()
```

Issue #231, opened 2023-03-20
Remove `useEagerConnect` hook


antho1404:
The `useEagerConnect` hook might not be necessary anymore as the wallet is auto-connected by wagmi now.
Still, the application requires the `ready` to query the API. We need to check if this `ready` is still relevant, if not then the whole `useEagerConnect` should be removed

antho1404:
merged with https://github.com/liteflow-labs/starter-kit/pull/291

PR #236, opened 2023-03-27
SSR improvements

antho1404:
Dependency: https://github.com/liteflow-labs/starter-kit/pull/241

### Description

Update SSR and use `getDataTree` to resolve all graphql queries
Improve the `useAccount` to have better compatibility on the server side
Fix cache issue with collection composed key

No more `getServerSideProps` 🎉 

### How to test

You can just run the different website pages with and without a refresh and check that the queries to get the data in the network panel do not have the pages' queries.
Also, you can use the settings `disable javascript` to really see the ssr.

### Checklist


client.ts
```diff
@@ -0,0 +1,35 @@
...
+    uri: environment.GRAPHQL_URL,
+    headers: authorization ? { authorization: `Bearer ${authorization}` } : {},
+    cache: new InMemoryCache({
+      typePolicies: {
+        Account: {
+          keyFields: ['address'],
+        },
+        Collection: {
+          keyFields: ['chainId', 'address'],
+        },
```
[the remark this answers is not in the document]
NicolasMahe: is it time to active the `nodeId` on the GraphQL schema?

hooks/useRequiredQueryParamSingle.ts
```diff
@@ -0,0 +1,18 @@
...
+    parse?: (value: string) => T
+  } = {},
+): T {
+  const value = useQueryParamSingle<T>(key, {
+    parse(value) {
+      invariant(value)
+      return options.parse ? options.parse(value) : (value as T)
+    },
+  })
+  invariant(value)
```
NicolasMahe, line 16: that's going to throw 500 errors when the url is malformed. should we use the `useQueryParamSingle` version and manage this in the pages?

pages/checkout/[id].tsx
```diff
@@ -28,80 +26,25 @@ import {
...
-
-const CheckoutPage: NextPage<Props> = ({
-  now,
-  offerId,
-  meta,
-  currentAccount,
-}) => {
-  const ready = useEagerConnect()
+const CheckoutPage: NextPage<Props> = ({ now }) => {
+  useEagerConnect()
```
Comment (NicolasMahe) [inline pages/checkout/[id].tsx:42]: Seems `useEagerConnect` is not doing anything. Could be removed.
```suggestion
```

Comment (NicolasMahe) [reply pages/checkout/[id].tsx:42]: Same comment for other pages using `useEagerConnect`

pages/create/index.tsx
```diff
@@ -75,20 +49,20 @@ const Layout = ({ children }: { children: React.ReactNode }) => (
...
         account: address || '',
         collectionFilter: collectionsFilter,
         fetchCollections: environment.MINTABLE_COLLECTIONS.length > 0,
       },
-      skip: !ready,
-    },
-  )
+    })
+
+  if (loading) return <></>
```
NicolasMahe, line 65: Why adding this special render on loading here? Is it breaking the render without it?

pages/_app.tsx
```diff
@@ -1,17 +1,13 @@
-import {
```
NicolasMahe, line 1: Is there a simple way to block the page from changing until it's content is loaded?

On almost every change of page, the website is blank until the queries are resolved.

https://user-images.githubusercontent.com/5823445/229420725-45ac2cc1-e30f-410d-a6c3-cda73254b751.mov

The experience of navigating the website is now pretty terrible because of this.

Only the page users bid, trades and offers seems to display placeholder data until its query are resolved, but they are missing a loader: https://nft-test-bsc-testnet-git-feature-ssr-improvement-liteflow.vercel.app/users/0x4b595014f7b45789c3f4e79324ae6d8090a6c8b5/bids

pages/tokens/[id]/bid.tsx
```diff
@@ -30,111 +28,34 @@ import {
...
-        },
-        currentAccount: ctx.user.address,
-      },
-    }
-  },
-)
-
-const BidPage: NextPage<Props> = ({ now, assetId, meta, currentAccount }) => {
-  const ready = useEagerConnect()
+const BidPage: NextPage<Props> = ({ now }) => {
```
Comment (NicolasMahe) [inline pages/tokens/[id]/bid.tsx:44]: There is an issue when I navigate to this page for the first time from another page.
It's working well after I refresh the page.

https://user-images.githubusercontent.com/5823445/229422462-9ac3717b-3e0b-4f14-9470-1613699aadc0.mov

NicolasMahe (commented):
That's much simpler change than I thought!
But there is a lot of work to do to make the experience as good as before, improving the loading state, better cache management and invalidation, etc...

client.ts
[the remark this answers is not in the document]
antho1404: Might help indeed but let's not do that right now

pages/checkout/[id].tsx
Comment (antho1404) [reply pages/checkout/[id].tsx:42]: Yes might be, but it's tracked here https://github.com/liteflow-labs/starter-kit/issues/231 so I didn't want to push on that as this requires dedicated testing which is not the goal of this PR

pages/create/index.tsx
antho1404: Yes if we don't, then data are not present and then we render the wrong information and we have flickering between wrong and right information which is worse than no information and information

pages/_app.tsx
antho1404: No, the only way is to properly handle loading states and/or to break the queries into smaller pieces (which is my goal). Still with loaders when required. I will work on loader on another PR(s) and merge into this one

pages/tokens/[id]/bid.tsx
Comment (antho1404) [reply pages/tokens/[id]/bid.tsx:44]: fixed in this PR as it is not really related to the SSR but a more general problem with forms that was highlighted here https://github.com/liteflow-labs/starter-kit/pull/241

pages/_app.tsx
```diff
@@ -287,4 +262,24 @@ function MyApp({
...
+
+MyApp.getInitialProps = async (
+  appContext: AppContext,
+): Promise<AppInitialProps<MyAppProps>> => {
+  const initialProps = await App.getInitialProps(appContext)
+  const cookies = (appContext.ctx.req as any)?.cookies // TODO: fix type
+  const jwt = cookies ? cookies['jwt-token'] : null
+  // Generate the now time, rounded to the second to avoid re-rendering on the client
+  // TOFIX: find a better way to share the time between the app and document
+  const now = Math.floor(Date.now() / 1000) * 1000
```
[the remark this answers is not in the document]
NicolasMahe: Did you try to inject the variable `now` in the document's `__NEXT_DATA__` like the `apolloState`?
It may be possible to access this variable from the app if it inside the `__NEXT_DATA__` in the right place.

pages/_document.tsx
```diff
@@ -19,6 +38,27 @@ class MyDocument extends Document {
       </Html>
     )
   }
+
+  static async getInitialProps(
+    context: DocumentContext,
+  ): Promise<DocumentInitialProps & { apolloState: NormalizedCacheObject }> {
+    invariant(context.req)
+    const cookies = (context.req as any).cookies // TODO: fix type
```
NicolasMahe, line 46: The `TODO: fix type` have been already fixed on the admin: https://github.com/liteflow-labs/admin/pull/82
Do you want to do it in another PR?

NicolasMahe (approved):
It's working pretty good but we now need to focus on improving the loading/flickering and invalidation of local cache.

pages/_app.tsx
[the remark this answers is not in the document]
antho1404: Yes, the issue here is that this is called twice on the server side, so not yet injected into the document.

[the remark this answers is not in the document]
antho1404: https://github.com/liteflow-labs/starter-kit/issues/250

pages/_document.tsx
antho1404: Yes I will do that in another PR to not block this one and add another review process

PR #241, opened 2023-04-03
Fix form data update on props change

antho1404:
### Description

Fix possible issues when a prop is updated, but the form is not updated as `useForm` is not refreshed and need a manual `useEffect` to refresh.

This fix does not have a massive impact currently but is an issue with the simplification of the SSR and the client side navigation as noticed here https://github.com/liteflow-labs/starter-kit/pull/236#discussion_r1155507068

### How to test

Navigate to the pages and check that data are updated correctly, on a parent component, implement a button or something to interact post render to see if the form data is correctly updated

### Checklist

- [x] Update docs if necessary


ismailToyran (approved):
Looking good 👍 
It feels bad that it doesn't actually get defaultValues on change, so need to use `useEffect` for that 😄

PR #244, opened 2023-04-10
Replace owned queries that use aggregation by asset.ownership

NicolasMahe:
### Description

Replace owned queries that use aggregation by a version that doesn't use aggregation.

### Checklist


antho1404 (approved):
This API is weird as we should always have only one item as we are already in an asset and we are filtering for one owner.
It could be great to have an API `owner(ownerAddress: $account)` that would already take into account the detail from the asset. And similar to the opposite relation with `account.owner(chainId, collection, tokenId)`.
If that is easily possible, that would be great.

NicolasMahe:
> This API is weird as we should always have only one item as we are already in an asset and we are filtering for one owner. It could be great to have an API `owner(ownerAddress: $account)` that would already take into account the detail from the asset. And similar to the opposite relation with `account.owner(chainId, collection, tokenId)`. If that is easily possible, that would be great.

I agree with you but as there is a solution without adding a new resolver, I was lazy to add it. I will do it now

PR #253, opened 2023-04-21
Add loader to all pages

antho1404:
Add missing loader to all pages.
Now navigation between pages displays a general loader.
With the current codebase and data fetching, this is the best solution, and the loading can be improved when splitting all the different queries into smaller ones, but this will require a lot more work and will probably be done page per page over time. Some page, like collection are already compatible because designed with multiple queries

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -13,6 +13,7 @@ import { NextPage } from 'next'
 import useTranslation from 'next-translate/useTranslation'
 import { useCallback, useEffect, useMemo } from 'react'
 import Link from '../components/Link/Link'
+import Loader from '../components/Loader'
 import Slider from '../components/Slider/Slider'
 import TokenCard from '../components/Token/Card'
 import TokenHeader from '../components/Token/Header'
@@ -58,7 +59,7 @@ const HomePage: NextPage<Props> = ({ now }) => {
     return (tokensToRender?.assets?.nodes || []).map((x) => x.id)
   }, [tokensToRender])
 
-  const { data, refetch, error } = useFetchHomePageQuery({
+  const { data, refetch, error, loading } = useFetchHomePageQuery({
     variables: {
       featuredIds: environment.FEATURED_TOKEN,
       now: date,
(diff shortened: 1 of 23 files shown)


pages/index.tsx
```diff
@@ -119,6 +120,9 @@ const HomePage: NextPage<Props> = ({ now }) => {
       )),
     [featured, address, signer, reloadInfo, currencies],
   )
+
+  if (loading) return <Loader fullPage />
```
NicolasMahe, line 124: The SSR on this page is timeout even on the demo: https://demo-bsc-testnet-git-feature-page-loaders-liteflow.vercel.app/
Seems to be the only page impacted.

NicolasMahe (approved):
SSR issue on the index but other page are working fine

ismailToyran (approved):
Looking good implementation wise.

I know this is just a start for client side implementation but I highly suggest we go SSR for user, asset and collection detail pages. We have all the data we need to show user a template for these detail pages. We could fetch client side rest of the data we need like total asset counts, likes, assets.

pages/tokens/[id]/index.tsx
```diff
@@ -315,11 +317,11 @@ const DetailPage: NextPage<Props> = ({ now: nowProp }) => {
             <Stack spacing={1}>
               {asset.collection.name && (
                 <Heading as="p" variant="heading1" color="gray.500">
-                  <Link
+                  <ChakraLink
```
Comment (ismailToyran) [inline pages/tokens/[id]/index.tsx:320]: https://github.com/liteflow-labs/starter-kit/pull/254 will override this to fix this issue application wise.

pages/index.tsx
ismailToyran: Not related with this PR
Checked deployments and this is the PR for that issue
But definetely needs to be fixed
https://github.com/liteflow-labs/starter-kit/commit/818f6d97f0e71ac676006af85886d7f0d03cf948#diff-aa98fd0757d0e1741503c50cfafb7726939d19819638bbe8e030a27adfec34a3

antho1404: indeed this is somehow platform-wise I will check on another PR

PR #255, opened 2023-04-25
Replace useChainId with useNetwork from wagmi

ismailToyran:
### Description
`useChainId` from wagmi was always returning chain id from the ones we have defined on chains.
`chain` from `useNetwork` doesn't need to be defined and always returns the correct chain Id whether that has been defined or not.

### How to test
Switch to other networks and use switch network button. It should work for all.

### Checklist


antho1404:
nice catch 👍

PR #258, opened 2023-04-25
Fix randomization of assets

### LINKED ISSUE #257: Issue with index page
The index page on the dev branch is not loading and has an timeout due to what seems to be an infinite loop

antho1404:
### Project organization

- Closes #257 

### Description

When the home page is loading with HOME_TOKENS variable set, the server-side rendering has an infinite loop because of the use of random that returns always a different array resulting in a never-ending fetching of assets with these new random ids

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -12,7 +12,6 @@ import { HiArrowNarrowRight } from '@react-icons/all-files/hi/HiArrowNarrowRight
 import { NextPage } from 'next'
 import useTranslation from 'next-translate/useTranslation'
 import { useCallback, useEffect, useMemo } from 'react'
-import invariant from 'ts-invariant'
 import Link from '../components/Link/Link'
 import Loader from '../components/Loader'
 import Slider from '../components/Slider/Slider'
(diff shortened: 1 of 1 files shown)


pages/index.tsx
```diff
@@ -51,13 +52,21 @@ const HomePage: NextPage<Props> = ({ now }) => {
...
+      const rng = (date.getTime() / 1000) % tokens.length
+      for (let i = 0; i < tokens.length; i++) {
+        const j = (i + rng) % tokens.length
+        const [temp1, temp2] = [tokens[i], tokens[j]]
+        invariant(temp1)
+        invariant(temp2)
+        tokens[i] = temp2
+        tokens[j] = temp1
+      }
+      return tokens.slice(0, environment.PAGINATION_LIMIT)
```
antho1404, line 66: @NicolasMahe, some confirmation would be great here, do you think about any other ways to pseudo-randomize this list of assets to ensure that every refresh does display a different list?

NicolasMahe: Not sure this logic is working as indeeded, I got some weird result when I run only this function in node.

Here is my suggested version:
```
const date = new Date()
const tokens = ['a', 'b', 'c', 'd']
console.log(tokens)

const seed = date.getTime()
console.log('seed', seed)

const randomTokens = []
while (tokens.length) {
  const random = seed % tokens.length
  const element = tokens.splice(random, 1)
  randomTokens.push(...element)
}

console.log(randomTokens)

```

If date is not in millisecond, then replace seed by `const seed = date.getTime() / 1000`.
Make sure to keep `const tokens = [...environment.HOME_TOKENS]` as the array token is modified.

antho1404: Good to go with that, a bit easier to understand and with the while loop we can also directly exit when we reach the pagination limit

Issue #273, opened 2023-04-29
Fix issue with collection detail page showing chains


antho1404:
When the application is configured with multichain, the collection page is displaying the assets filter including the chain filter which doesn't make sense as we are already in a collection attached to a chain

antho1404:
fixed by https://github.com/liteflow-labs/starter-kit/pull/275

PR #275, opened 2023-04-29
disable chain filter on the collection page

antho1404:
Closes https://github.com/liteflow-labs/starter-kit/issues/273

Fix issue with collection detail page that displayed the chain filter and the collection on mobile


ismailToyran (approved):
Looking good 👍

PR #276, opened 2023-04-30
Add emails notifications for bid/auction/offer/trades

antho1404:
### Description

Add the list of all emails in the platform to be connected with webhook.
Thanks to this, emails can now be fully customized by the client in terms of text, flow, or design.
A few tools can be used now to create beautiful emails like:
- https://react.email/
- https://mjml.io/
- or any email builder

### How to test

Add all the webhooks in the admin.
Screenshot 2023-04-30 at 23 52 10

If tested locally, make sure that the URL is public using a tool like ngrok or localtunnel

### DIFF
--- a/emails/AuctionEndedNoBids.ts
+++ b/emails/AuctionEndedNoBids.ts
@@ -1,0 +1,39 @@
+import { formatDate } from '@nft/hooks'
+import { Events } from '@nft/webhook'
+import environment from '../environment'
+import { fetchAdditionalAuctionData } from './auction-utils'
+
+export default async function AuctionEndedNoBids({
+  id,
+  endAt,
+  asset,
+  creator,
+}: Events['AUCTION_ENDED']): Promise<{
+  html: string
+  subject: string
+  to: string
+} | null> {
+  const { expireAt, bestOffer } = await fetchAdditionalAuctionData(id, endAt)
+
+  if (!!bestOffer) return null
+  if (!creator?.email) return null
+  return {
+    to: creator.email,
+    subject: `Your auction has ended but no bids were placed for ${asset.name}`,
+    html: `Hi <strong>${creator.username || creator.address}</strong>,<br/>
+    <br/>
+    Your auction for the NFT <strong>${
+      asset.name
+    }</strong> has ended but hasn't received any bids.<br/>
+    <br/>
+    Friendly tips, create a new sale for this NFT and share it around you to gather interest.<br/>
+    <br/>
+    Without action from your side after the <strong>${formatDate(
+      expireAt,
+    )}</strong>, the auction and its bids will be canceled.<br/>
+    <br/>
+    <a href="${environment.BASE_URL}/tokens/${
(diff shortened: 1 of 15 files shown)


emails/auction-utils.ts
```diff
@@ -0,0 +1,92 @@
...
+    query,
+    {
+      id: auctionId,
+      now: new Date(endAt),
+    },
+    // TODO: fix this request as `email` cannot be fetched on the public API
+    // and because of that emails to bid creators are not sent
+    // {
+    //   Authorization: `Bearer ${process.env.LITEFLOW_ADMIN_TOKEN}`,
+    // },
```
antho1404, line 76: @NicolasMahe, any idea how to solve that?
Either we need a new type of authentication for api keys or we could just send this data directly in the payload of the webhook, this way this whole fetching would not be required.
What do you think?

NicolasMahe: Yes let's add any data that can be useful directly in the webhook's payload to avoid having to query the API here.

antho1404: Added some work on the infrastructure to expose these additional data and added a dependency to this PR to wait for this

antho1404:
@NicolasMahe could be great to have a second validation on this one to check that I didn't miss any emails nor any conditions to trigger these emails

emails/AuctionEndedNoBids.ts
```diff
@@ -0,0 +1,36 @@
...
+    <br/>
+    Your auction for the NFT <strong>${
+      asset.name
+    }</strong> has ended but hasn't received any bids.<br/>
+    <br/>
+    Friendly tips, create a new sale for this NFT and share it around you to gather interest.<br/>
+    <br/>
+    Without action from your side after the <strong>${formatDate(
+      expireAt,
+    )}</strong>, the auction and its bids will be canceled.<br/>
```
NicolasMahe, line 30: We have the same misleading text on the backend, but let's fix it here:
```suggestion
    )}</strong>, the auction will be canceled.<br/>
```

pages/api/notification.ts
```diff
@@ -26,31 +37,42 @@ const transporter = nodemailer.createTransport({
 
 const emails = new Map<
   keyof Events,
-  ((data: Events[keyof Events]) => SendMailOptions | null)[]
+  ((data: any) => SendMailOptions | null | Promise<SendMailOptions | null>)[]
```
NicolasMahe, line 40: I don't think any email function need to be async anymore:
```suggestion
  ((data: any) => SendMailOptions | null>)[]
```

NicolasMahe (approved):

pages/api/notification.ts
antho1404: indeed this was required because the data were not in the webhook but this is now solved so I simplified the logic [c4f5360](https://github.com/liteflow-labs/starter-kit/pull/276/commits/c4f53604010f2758c2ba25faaa4fc5da6e4878b3)

Issue #217, opened 2023-03-03
Display user profile in the bid/offers/trades


antho1404:
The tables in the user profile with the bids/offers/trades display only the address of the user and not its name and verification badge if present. We should display all the information when possible

PR #292, opened 2023-05-09
Display user profile on activity

antho1404:
Closes https://github.com/liteflow-labs/starter-kit/issues/217

On the user page, when listing the bids and trades, we were only displaying the address regardless to the fact that the buyer/seller/bidder may have a profile with a name and image

| Before | After | 
| --- | --- |
| Screenshot 2023-05-09 at 23 30 42 | Screenshot 2023-05-09 at 23 30 28 |

@EmmanuelDrouin are we good to merge this or is there any reason not to display accounts?


EmmanuelDrouin:
@antho1404 good to merge 👍

Issue #221, opened 2023-03-07
New values for locale or remove locale for traits


ismailToyran:
We should either add all traits that are belong to category to the translate folder or use the ones that come from API only.
Using fallback still throws annoying none stop warning to the console

![image](https://user-images.githubusercontent.com/33982052/223360047-b6db1450-6bd9-4f88-a361-006e411e998c.png)

antho1404:
fixed by https://github.com/liteflow-labs/starter-kit/pull/300

Issue #77, opened 2022-10-05
Improve flickering when account changes, login, or logout


NicolasMahe:
Since https://github.com/liteflow-labs/nft/pull/2075, when the user login or logout on one the following pages, the UI is rendered with an intermediate blank state:

https://user-images.githubusercontent.com/5823445/193965886-f34051b2-e871-4d77-a03b-6c94dee9559f.mov

Here are the pages in template when this issue occurs (always because the query is using a variable with the account's address):
- [ ] `useFetchAssetQuery` in `packages/templates/src/asset/Detail.tsx`
- [ ] `useFetchCategoriesAndAccountQuery` in `packages/templates/src/asset/Form.tsx`
- [ ] `useFetchHomePageQuery` in `packages/templates/src/home/Home.tsx`
- [ ] `useOfferForAssetQuery` in `packages/templates/src/offers/Form.tsx`
- [ ] `useFetchAccountVerificationStatusQuery` in https://github.com/liteflow-labs/nft/pull/2123/files#diff-25c058c98646a585ec1e6d205a486ae96318511366406dba4c828481e4b37293R55 (also missing fallback to account from server)

---

I investigate for a solution and I find out that Apollo set the data of the query to undefined as soon as it starts a new fetch.
This cause the UI to re-render but without data. When Apollo gets the data, a new render is executed with the data.
This is a change in Apollo from v2 to v3 and some people complain: https://github.com/apollographql/apollo-client/issues/6866

Apollo quickly introduced after this breaking change a new variable `previousData` that contains the data from the previous fetch if any to let the developer access previous data.

The simple solution to fix this issue is to add `previousData` as default of `data`.

Here is an example of this solution on `packages/templates/src/asset/Detail.tsx`
```diff
 const {
-  data
+  previousData,
+  data = previousData, // fallback to previousData when the query is loading
   refetch,
 } = useFetchAssetQuery({
   variables: {
     id: assetId,
     now: date,
     address: (ready ? account?.toLowerCase() : currentAccount) || '',
   },
 })
```

Here is the result with this change:

https://user-images.githubusercontent.com/5823445/193966038-c3874e22-ddcd-47b1-ad22-d8a9ed621b04.mov

Even better, we can display the `NProgress` loading animation while the query is loading:
```diff
 const {
-  data
+  previousData,
+  data = previousData, // fallback to previousData when the query is loading
   refetch,
+  loading,
 } = useFetchAssetQuery({
   variables: {
     id: assetId,
[...]

Issue #63, opened 2022-08-23
Deactivate Apollo client cache


NicolasMahe:
I would like to open the discussion about completely removing the cache of Apollo.

Currently, the codebase doesn't properly implement the cache system of Apollo.

Only type `Account` is setup to merge data with the same `address` and I'm not even sure why.

Apollo seems to cache by default lazy query (bug because of cache on lazy query: https://github.com/liteflow-labs/nft/pull/1948) but not on all version.

My suggestion is to deactivate the cache of Apollo for now:
```diff
return new ApolloClient({
    link: authLink.concat(httpLink),
    cache: new InMemoryCache({
      typePolicies: {
        Account: {
          keyFields: ['address'],
        },
      },
    }),
    ssrMode,
+    defaultOptions: {
+      query: {
+        fetchPolicy: 'no-cache',
+      },
+      mutate: {
+        fetchPolicy: 'no-cache',
+      },
+      watchQuery: {
+        fetchPolicy: 'no-cache',
+      },
    },
  })
```
https://github.com/liteflow-labs/nft/blob/4ebf01b1a77741542576296efdbd7e70a0081f6f/packages/components/src/client.ts#L51-L61

This will help remove any side-effect caused by un-wanting cache.

We can always activate the cache per request if we need to.

@liteflow-labs/nft-marketplace I would like to have your feedback on this one.

EmmanuelDrouin:
@antho1404 could you elaborate on the "effort" required for this? Chris is mentioning a "big refactor" I would like to understand the effort on our side to take a decision on this issue ;)

antho1404:
This still needs more research to see if it even makes sense to remove the cache because of the SSR. For information, I have already started to work on some improvements on the SSR and the apollo client/cache here https://github.com/liteflow-labs/starter-kit/pull/118. This will give more answers

antho1404:
fixed by https://github.com/liteflow-labs/starter-kit/pull/300

PR #300, opened 2023-05-22
Fix issues with SSR and improve cache and networking

antho1404:
### Project organization

- Closes https://github.com/liteflow-labs/starter-kit/issues/77
- Closes https://github.com/liteflow-labs/starter-kit/issues/63
- Closes https://github.com/liteflow-labs/starter-kit/issues/221

### Description

- Fix a few issues with the Apollo Cache because of the change of id for the assets
- Force to read from the cache, then still refetch in the background with the `cache-and-network` option
- Fixed a couple of markup errors that where triggering some invalid render between server and client
  - div in p, a in a, p in a, hr in p...
- Update skeleton to not have them when the page is initially loaded (and data is re-fetched in the background)
  - Pages with "fixed" data that do not involve any update of data in the page simply display the skeleton based on the presence of data
  - Pages with "dynamic" data that can be filtered/sorted display the skeleton based on the loading state (so during the initial load with the refetch)

### Todo

~~- Improve loading when switching filters in the account pages currently treated as "fixed" data instead of "dynamic"~~
~~- Decide on the best solution between the fixed/dynamic data with either `loading || !data` or `loading && !data`~~
https://github.com/liteflow-labs/starter-kit/issues/302


components/Navbar/Navbar.tsx
```diff
@@ -545,12 +546,9 @@ const Navbar: VFC<{
...
           ) : (
-            <ConnectButton
-              accountStatus="address"
-              chainStatus="none"
-              showBalance={false}
-              label={t('navbar.sign-in')}
-            />
+            <Button onClick={openConnectModal} colorScheme="brand">
+              {t('navbar.sign-in')}
+            </Button>
```
ismailToyran, line 551: Best change ever, didn't know they had a hook for that 🎉

ismailToyran (approved):
Looking great 👍 
Didn't find anything blocking. 🎉

PR #305, opened 2023-05-28
Disable unused top up option

antho1404:
Remove the option to top up your account with fiat, this is an option that has never been used and that needs to manually be removed by everyone using the starter kit.


antho1404:
close in favor of https://github.com/liteflow-labs/starter-kit/pull/306

PR #310, opened 2023-06-05
New environment file

antho1404:
Simplify the configuration of the starter kit by reducing the list of environmental variables required and better documenting the `./environment.ts` file

@EmmanuelDrouin I would love your feedback on this file to see if it's easy to understand https://github.com/liteflow-labs/starter-kit/blob/15684efedac9be81baea06a7b82a725ff9a94cde/environment.ts


environment.ts
```diff
@@ -1,116 +1,114 @@
...
-  .filter(Boolean)
-  .map((address) => ({ address: address.toLowerCase(), chainId: CHAIN_ID }))
+  // API Key for the Liteflow API, you can get one at https://dashboard.liteflow.com/developer
+  LITEFLOW_API_KEY: process.env.NEXT_PUBLIC_LITEFLOW_API_KEY,
 
-const environment: Environment = {
-  MAGIC_API_KEY: process.env.NEXT_PUBLIC_MAGIC_API_KEY,
-  GRAPHQL_URL: process.env.NEXT_PUBLIC_GRAPHQL_URL,
-  FEATURED_TOKEN: process.env.NEXT_PUBLIC_FEATURED_TOKEN.split(','),
+  // Email address to send reports to
```
EmmanuelDrouin, line 22: @antho1404 should we specify "for end users to send reports to"?
I'm wondering if as a first step (you configure this file) you might get stuck 2 min on this one to fully understand this report functionality

environment.ts
```diff
@@ -1,116 +1,114 @@
...
-  CHAIN_IDS: CHAIN_IDS,
-  REPORT_EMAIL: process.env.NEXT_PUBLIC_REPORT_EMAIL,
-  HOME_COLLECTIONS: process.env.NEXT_PUBLIC_HOME_COLLECTIONS?.split(','),
-  HOME_TOKENS: process.env.NEXT_PUBLIC_HOME_TOKENS?.split(','),
-  HOME_USERS: process.env.NEXT_PUBLIC_HOME_USERS?.split(','),
-  OFFER_VALIDITY_IN_SECONDS: OFFER_VALIDITY_IN_SECONDS,
-  AUCTION_VALIDITY_IN_SECONDS: AUCTION_VALIDITY_IN_SECONDS,
-  BUGSNAG_API_KEY: process.env.NEXT_PUBLIC_BUGSNAG_API_KEY,
+
+  // Number of seconds an offer is valid (recommendation: 28 days)
```
EmmanuelDrouin, line 28: @antho1404 is this part about the "maximum time" a user can set an offer? In your example it means users will not be able to create offers valid for more than 28 days, right?

If it's the maximum a user can set, we should specify this with "Maximum number of seconds an offer can be valid" as currently it can be understood as "all users will have to set an offer valid for 28 days"

EmmanuelDrouin: Should we talk about expiration if this is related to the offer expiring? 
> Maximum number of seconds a user can set before an offer expired

environment.ts
```diff
@@ -1,116 +1,114 @@
...
+  OFFER_VALIDITY_IN_SECONDS: 2419200, // 28 days
+
+  // Number of seconds an auction is valid (recommendation: 7 days)
```
EmmanuelDrouin, line 31: Similar comment here as the one for the offers above

environment.ts
```diff
@@ -1,116 +1,114 @@
...
+  AUCTION_VALIDITY_IN_SECONDS: 604800, // 7 days
+
+  // Base URL of the website
   BASE_URL: process.env.NEXT_PUBLIC_BASE_URL,
-  UPLOAD_URL: process.env.NEXT_PUBLIC_UPLOAD_URL,
-  REFERRAL_PERCENTAGE: { base: 20 * 0.025, secondary: 20 * 0.01 },
-  RESTRICT_TO_VERIFIED_ACCOUNT: true,
+
+  // Maximum amount of royalties
```
EmmanuelDrouin, line 37:
```suggestion
  // Maximum percentage of royalties
```

environment.ts
```diff
@@ -1,116 +1,114 @@
...
+  // List of tokens randomized to be featured on the homepage with the following format: [chainId]-[contractAddress]-[tokenId]
+  // If empty, the tokens will be the last created ones
+  HOME_TOKENS: [],
+
+  /**
+   * Wallet/chain configuration
+   */
+
+  // List of supported chains. Liteflow is supporting the following: mainnet, goerli, bscTestnet, bsc, polygon, polygonMumbai
+  CHAINS: [mainnet, goerli, bscTestnet, bsc, polygon, polygonMumbai],
```
EmmanuelDrouin, line 83: @antho1404 is there anyway we could rename "Mainnet" for Ethereum? I understand it but this will confuse users

environment.ts
```diff
@@ -1,116 +1,114 @@
...
+
+  // (Optional) Magic API Key, you can get one at https://magic.link/
+  MAGIC_API_KEY: process.env.NEXT_PUBLIC_MAGIC_API_KEY,
+
+  /**
+   * SEO Configuration
+   */
+
+  // Name of the company to place in the footer and in the SEO title
```
EmmanuelDrouin, line 92:
```suggestion
  // Name of the company to place in the SEO title and in the footer 
```

environment.ts
```diff
@@ -1,116 +1,114 @@
...
+
+  // Title of the marketplace to place in the SEO title
+  META_TITLE: 'Acme NFT Marketplace',
+
+  // Description of the marketplace to place in the SEO description
+  META_DESCRIPTION: 'Acme NFT Marketplace',
+
+  // Keywords of the marketplace to place in the SEO keywords
+  META_KEYWORDS:
+    'NFT, marketplace, platform, white-label, blockchain, liteflow',
```
EmmanuelDrouin, line 103:
```suggestion
    'NFT, marketplace, platform, blockchain, liteflow',
```

environment.ts
```diff
@@ -1,116 +1,114 @@
...
+
+  /**
+   * Miscellaneous
```
EmmanuelDrouin, line 106: @antho1404 could you add one line of context for each point in this section?

Currently a user will not understand that the "RESTRICT_TO_VERIFIED_ACCOUNT" is for the mint, etc

EmmanuelDrouin (commented):
@antho1404 overall really good, it is super clear and easy to understand and configure good job 👍 

I only left a few comments to improve the understanding on some part from a new user point of view but apart from that it rocks 🔥

environment.ts
antho1404: This is related to the default value, the user will be able to change anything, I will add that information

environment.ts
antho1404: Yes this one is ugly on purpose as this is removed by other PRs that removes these configs

environment.ts
antho1404: done

environment.ts
antho1404: done

environment.ts
antho1404: renamed into ethereumMainnet and ethereumGoerli

environment.ts
antho1404: done

environment.ts
antho1404: added "Default value for the number ..."

environment.ts
antho1404: done

environment.ts
antho1404: done

PR #311, opened 2023-06-05
Remove config for mint

antho1404:

Instead of having an option to disable the minting, if an application does not require minting then they can simply remove the links as they will anyway have to remove the page as well


antho1404:

NicolasMahe (approved):
all good 👌

EmmanuelDrouin (approved):

PR #312, opened 2023-06-05
Remove banner

antho1404:

Remove the Liteflow banner to show it is a demo. This is only needed for the demo, which has moved to a dedicated repository https://github.com/liteflow-labs/demo.
This is a customization for Liteflow and should not be part of the starter kit


antho1404:

NicolasMahe (approved):
all good 👌

EmmanuelDrouin (approved):

PR #314, opened 2023-06-05
Remove restriction on mint

antho1404:

Remove the restriction on mint, as this is a business decision and should not be part of the starter kit. Clients can reimplement it their own ways if they need to


antho1404:

NicolasMahe (approved):
all good 👌

EmmanuelDrouin (approved):

PR #315, opened 2023-06-05
Fetch mintable collections instead of getting from env

antho1404:
Dependency: https://github.com/liteflow-labs/starter-kit/pull/311
Dependency: https://github.com/liteflow-labs/starter-kit/pull/314

Mintable collection can be fetched from the infra but required to have an env variable to define them which could lead to inconsistency.


antho1404:

NicolasMahe (approved):
all good 👌

EmmanuelDrouin (approved):

PR #318, opened 2023-06-21
Moving utils from sdk to custom utils

antho1404:
Move all the updates directly to the starter kit to prepare for the sdk v1 release


hooks/useBalance.ts
```diff
@@ -0,0 +1,20 @@
+import { getAddress } from '@ethersproject/address'
+import { BigNumber } from '@ethersproject/bignumber'
+import { useBalance as useWagmiBalance } from 'wagmi'
+
+export default function useBalance(
```
antho1404, line 5: New hook to keep the same API we had just using wagmi instead of the SDK that has this function deprecated

utils.ts
```diff
@@ -0,0 +1,70 @@
+import { getAddress, isAddress } from '@ethersproject/address'
```
antho1404, line 1: utils copied from https://github.com/liteflow-labs/liteflow-js/tree/0b8ad9530e09f60081e040701f925a1473d2eb8f/packages/hooks/src/utils

PR #323, opened 2023-06-26
Split total count query in explorer page

NicolasMahe:
### Project organization

- Dependant on https://github.com/liteflow-labs/starter-kit/pull/322

### Description

Split total count query in explorer page


pages/explore/explore.tsx
```diff
@@ -70,6 +74,14 @@ const ExplorePage: NextPage<Props> = ({ now }) => {
     },
```
NicolasMahe, line 74: @antho1404 can you confirm why there is a `previousData` + a `loading` on `useFetchAllErc721And1155Query`?
It seems the data are not displayed if `loading = true`, thus making `previousData` pretty much useless.
Is it something due to SSR?

pages/explore/explore.tsx
```diff
@@ -70,6 +74,14 @@ const ExplorePage: NextPage<Props> = ({ now }) => {
...
   })
 
+  const { data: totalCountData } = useFetchAllErc721And1155TotalCountQuery({
+    variables: {
+      filter: convertFilterToAssetFilter(filter, date),
+    },
+    ssr: false,
```
NicolasMahe, line 81: Not executing this on SSR has the bonus of not sending and rendering the thousands and thousands of select option to select the exact page. The initial page load is only 25Kb vs 75Kb and take less than 1sec vs 3sec.

pages/explore/explore.tsx
```diff
@@ -70,6 +74,14 @@ const ExplorePage: NextPage<Props> = ({ now }) => {
     },
   })
 
+  const { data: totalCountData } = useFetchAllErc721And1155TotalCountQuery({
```
NicolasMahe, line 77: I didn't implement a loading state here, so the element is completely hidden while loading.
Could be good to add a skeleton loader.

NicolasMahe: @ismailToyran how long will it take to add this skeleton?

ismailToyran: Should take around 30 to 60min, considering grid + pagination.

pages/explore/explore.tsx
NicolasMahe: closing in favour of https://github.com/liteflow-labs/starter-kit/pull/328

antho1404 (approved):

pages/explore/explore.tsx
antho1404: Not a big fan of having a specific behavior just for that query, that creates inconsistency in the codebase but I will not block for this

pages/explore/explore.tsx
NicolasMahe: created dedicated issue to not block this PR: https://github.com/liteflow-labs/starter-kit/issues/340

pages/explore/explore.tsx
NicolasMahe: Indeed, this could be added to other queries with total count. The bigger the platform are becoming, the more queries will be slow because of the total count.

Issue #302, opened 2023-05-23
Improve skeleton for pages with filters/sort


antho1404:
Pages with filter/sorts have the skeleton, but because we need the skeleton, the initial loading with https://github.com/liteflow-labs/starter-kit/pull/300 also displays the skeleton even tho data are already available.

The explore pages have the skeleton when loading, filtering, and sorting, while the account pages don't have these skeletons resulting in an initial load with no skeleton.

We need to normalize this and, even better, find a way that could "disable" the skeleton for the first load but not after.
Maybe some things with `initialFetchPolicy`, `nextFetchPolicy`, maybe delaying the `cache-and-network` only after the component is mounted with a `useEffect`

NicolasMahe:
This is fixed by https://github.com/liteflow-labs/starter-kit/pull/328

PR #328, opened 2023-06-27
Do not display skeleton display when data exist

NicolasMahe:
### Project organization

- Closes https://github.com/liteflow-labs/starter-kit/issues/302

### Description

Fix issue when skeleton was display (and in a weird way) on initial page load.
As you can see on the following video, the initial SSR works fine, then the skeleton is display but with a height of zero, then the client replace the skeleton with the right data:

Before:

https://github.com/liteflow-labs/starter-kit/assets/5823445/a3ede65b-28d3-41d4-a4f0-ed6e3c0eb33c

After:

https://github.com/liteflow-labs/starter-kit/assets/5823445/a2dbe99c-7abc-4649-8b0c-16b83ecaee4e

Changing the logic of skeleton display(eg on explore page, `loading || !assetsData` to `loading && !assetsData`) prevents the skeleton to display on filter changes when `assetsData` is also initialised with `previousData`. So I also remove the following piece of code on the impacted pages:
```ts
const assetsData = useMemo(() => data || previousData, [data, previousData])
```

This changes also allow to display the data as soon as we have them, and let Apollo refetch the query in the background because of the `cache-and-network` policy.

With data already loaded
Before:

https://github.com/liteflow-labs/starter-kit/assets/5823445/c75bbfd0-fc21-4e01-9eb2-3d68647b591f

After:

https://github.com/liteflow-labs/starter-kit/assets/5823445/e9ca94f8-8fc8-414e-bcf3-e6a0e7a9c2de

### Checklist


pages/explore/explore.tsx
```diff
@@ -169,7 +167,7 @@ const ExplorePage: NextPage<Props> = ({ now }) => {
               </GridItem>
             )}
             <GridItem gap={6} colSpan={showFilters ? 1 : 2}>
-              {loading || !assetsData ? (
+              {loading && !assetsData ? (
```
NicolasMahe, line 170: Note for reviewer: I applied the modifications on all pages that were using `loading || !assetsData`: checkout, collection, explore/collection, explore/users

pages/checkout/[id].tsx
```diff
@@ -52,10 +52,7 @@ const CheckoutPage: NextPage<Props> = ({ now }) => {
   const date = useMemo(() => new Date(now), [now])
   const offerQuery = useCheckoutQuery({ variables: { id: offerId } })
 
-  const offer = useMemo(
-    () => offerQuery.data?.offer || offerQuery.previousData?.offer,
-    [offerQuery.data, offerQuery.previousData],
-  )
+  const offer = offerQuery.data?.offer
```
Comment (NicolasMahe) [inline pages/checkout/[id].tsx:55]: I checked the codebase and most (if not all) queries are using the `previousData` logic.
@antho1404 any reason to have implemented it on all queries?
Removing the `previousData` logic feel a lot nicer in term of UX. The skeletons are displayed when queries are running instead of having no feedback until the queries are resolved.

NicolasMahe:
PR https://github.com/liteflow-labs/starter-kit/pull/332 make the explore page to infinite load because of the weird `loading || !assetsData` that this new PR fixes:
Without this PR: https://nft-test-uaan-4w6ulcoit-liteflow.vercel.app/explore
With this PR: https://nft-test-uaan-git-fix-skeleton-display-when-data-exist-liteflow.vercel.app/explore

pages/checkout/[id].tsx
Comment (antho1404) [reply pages/checkout/[id].tsx:55]: This was needed because the initial load was fetching the data, and that was leading to a page fully loaded that "reloads". This has been fixed by https://github.com/liteflow-labs/starter-kit/pull/332 but somehow when I did the migration of that was not working (maybe a version mismatch) thus the need to implement that

antho1404 (approved):

PR #329, opened 2023-06-29
Fix skip logic in component HomeSection Collections and Users

NicolasMahe:
### Description

I notice that the homepage on hodooi.com is executing the query `FetchCollectionsQuery` without any collections and took 1min to load because the database is busy.
The `skip` logic was not implemented correctly. An empty or filled array is always truly.

### Checklist


ismailToyran:
Oops nice catch, thanks 👍

ismailToyran (approved):

PR #334, opened 2023-07-02
Update robots.txt

NicolasMahe:
### Project organization

- Related https://trello.com/c/zuxaXXkh/919-bots-issue-investigation

### Description

Remove wrong disallow rules and add crawl delay in robots.txt

### Checklist

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,2 +1,2 @@
 User-agent: *
-Disallow:
+Crawl-Delay: 10


public/robots.txt
```diff
@@ -1,2 +1,2 @@
 User-agent: *
-Disallow:
```
NicolasMahe, line 2: @antho1404 what was the goal of this `Disallow`? Was it to block all bots from indexing the website?
The correct rule tell bots not to index is:
```
Disallow: /
```

antho1404: No, it was to disallow nothing. It was probably never configured and used a default template

PR #337, opened 2023-07-05
Set minimumCacheTTL to 1 year in next image config

NicolasMahe:
### Description

Set minimumCacheTTL to 1 year in next image config

### Checklist


antho1404:
Not a massive issue but this could create an issue with dNFTs and reveal as the url of the image will not change but its content will

PR #345, opened 2023-07-20
Normalize filters in graphql

ismailToyran:
### Project organization
- Part of the [trello card](https://trello.com/c/gsVDdQpO/930-starter-kit-miscellaneous-issues-step-2-3-4)

### Description
Normalizes filters across the application

### Checklist

### DIFF
--- a/hooks/useCollectionFilterFromQuery.ts
+++ b/hooks/useCollectionFilterFromQuery.ts
@@ -24,12 +24,13 @@ const searchFilter = (search: string): CollectionFilter =>
 
 export const convertFilterToCollectionFilter = (
   filter: Filter,
-): CollectionFilter[] => {
-  const queryFilter = []
-  if (filter.chains && filter.chains.length > 0)
-    queryFilter.push(chainFilter(filter.chains))
-  if (filter.search) queryFilter.push(searchFilter(filter.search))
-  return queryFilter
+): CollectionFilter => {
+  return {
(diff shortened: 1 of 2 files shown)


hooks/useCollectionFilterFromQuery.ts
```diff
@@ -24,12 +24,13 @@ const searchFilter = (search: string): CollectionFilter =>
...
-  if (filter.chains && filter.chains.length > 0)
-    queryFilter.push(chainFilter(filter.chains))
-  if (filter.search) queryFilter.push(searchFilter(filter.search))
-  return queryFilter
+): CollectionFilter => {
+  return {
+    ...(filter.chains && filter.chains.length > 0
+      ? chainFilter(filter.chains)
+      : {}),
+    ...searchFilter(filter.search || ''),
```
ismailToyran, line 32: As we are using object instead of an array, we need to have at least one field present.
So I've forced search filter to always have a value, in this case `''`
Not the best solution but if we want to remove `{ and: $filter }` usage, we can not return `{}` anymore.
What do you think @antho1404

antho1404: humin that case maybe let's keep it as it was, I'm afraid this might have more side effects than benefits

ismailToyran:
Closing after resolving conversation. This is not needed.

PR #351, opened 2023-07-27
Add optional Alchemy provider as fallback when public providers are not responsive

NicolasMahe:
### Description

Add optional Alchemy provider as fallback when public providers are not responsive.
It's based on a new env `NEXT_PUBLIC_ALCHEMY_API_KEY`

### How to test

- With the dev/main branch, open your network console, go to wallet page, notice the error to `https://matic-mumbai.chainstacklabs.com/` and the Mumbai's token with zero of balances
- With this branch, open your network console, go to wallet page, notice the error to `https://matic-mumbai.chainstacklabs.com/`, then see the working request to `https://polygon-mumbai.g.alchemy.com/v2/wtMo0BLwnWWHGWBi-WpVc9d9pFcVPMW3`, Mumbai's token should have the correct balance

### Checklist

### To do before merge

- [x] Add env `NEXT_PUBLIC_ALCHEMY_API_KEY` to uaan


connectors.ts
```diff
@@ -23,10 +23,16 @@ import {
 } from 'wagmi/chains'
```
NicolasMahe, line 23: @antho1404 I just would like to confirm this direction with you. Already discussed with Manu on https://github.com/liteflow-labs/o-mee/issues/702

antho1404: For information, I already created an issue on wagmi related to that https://github.com/wagmi-dev/references/issues/449. I suggest waiting for a positive answer for a few days and merging this if nothing.

Also, instead of having a global alchemy provider, I would have just changed the default rpc of the mumbai chain, this way, it only uses a custom alchemy provider for mumbai, but that depends on how "permanent" we want this fix.

NicolasMahe: The alchemy provider is added second, so as a fallback. If you go to your wallet page you will see that all other chains are using the default providers.
When wagmi fixes its default provider, then Alchemy will not be needed.
I think this feature is very good to not bother users with faulty default providers.

antho1404 (approved):

connectors.ts
antho1404: This is optional anyway, so yes, that's not bad to add. Hopefully, it shouldn't be needed, but we can have that for now, even if I believe we might remove it soon when wagmi fixes it to simplify the starter kit. This should be a business decision to add alchemy, and we shouldn't dictate this

PR #352, opened 2023-07-27
remove fallback to 0 in useBalance and WalletBalance

NicolasMahe:
### Description

Remove fallback to `0` in `useBalance` and `WalletBalance`. Now it display `-` if the balance was not properly loaded.

Before:
Screenshot 2023-07-27 at 15 04 02

After:
Screenshot 2023-07-27 at 15 03 28

### Checklist

### DIFF
--- a/components/Wallet/Account/WalletBalance.tsx
+++ b/components/Wallet/Account/WalletBalance.tsx
@@ -22,7 +22,8 @@ const WalletBalance: FC<{
         speed="0.65s"
       />
     )
-  return <Price amount={balance || 0} currency={currency} />
+  if (!balance) return '-'
+  return <Price amount={balance} currency={currency} />
 }
 
 export default WalletBalance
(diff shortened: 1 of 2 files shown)


components/Wallet/Account/WalletBalance.tsx
```diff
@@ -22,7 +22,8 @@ const WalletBalance: FC<{
         speed="0.65s"
       />
     )
-  return <Price amount={balance || 0} currency={currency} />
+  if (!balance) return '-'
```
antho1404, line 25: Be careful if the balance is 0, then we will display `-`. I get that if there is an error or something like that, we don't want to display 0, but if there is no error, I believe it's better to display 0 instead of `-` that is "unknown"

NicolasMahe: The type of `balance` is `BigNumber | undefined`, so this condition will return true only when balance is undefined.

Screenshot 2023-07-28 at 10 27 39

antho1404: My bad, perfect then 😀

PR #367, opened 2023-08-14
Simplify pagination and change usage

ismailToyran:
### Project organization
- Related [trello card](https://trello.com/c/VLhWOSFG/1000-starter-kit-perf-issues)

### Description
Fixes pagination slow query issues

### How to test
Use any of the paginations on the platform

### Checklist

### DIFF
--- a/components/Pagination/Pagination.tsx
+++ b/components/Pagination/Pagination.tsx
@@ -35,7 +35,7 @@ export default function Pagination({
   onPageChange,
   ...props
 }: IProp) {
-  const { t } = useTranslation('templates')
+  const { t } = useTranslation('components')
 
   if (hasPreviousPage === undefined || hasNextPage === undefined)
     return (
@@ -59,7 +59,7 @@ export default function Pagination({
       {!hideSelectors && (
         <Flex gap={3} display={{ base: 'none', sm: 'flex' }}>
           <HStack spacing={1} minWidth="max">
-            <FormLabel m={0}>{t('pagination.result.label')}</FormLabel>
+            <FormLabel m={0}>{t('pagination.label')}</FormLabel>
           </HStack>
           <Select
             cursor="pointer"
--- a/components/Pagination/Pagination.tsx
+++ b/components/Pagination/Pagination.tsx
@@ -14,24 +14,26 @@ import { IoChevronForward } from '@react-icons/all-files/io5/IoChevronForward'
 import useTranslation from 'next-translate/useTranslation'
 
 export type IProp = {
-  limit: number
-  limits?: number[]
   page: number
   hasNextPage: boolean | undefined
   hasPreviousPage: boolean | undefined
-  hideSelectors?: boolean
-  onLimitChange?: (limit: string) => void
   onPageChange: (page: number) => void
-}
+} & (
+  | {
+      withoutLimit: true
+    }
+  | {
+      withoutLimit?: false
+      limit: number
+      limits: number[]
+      onLimitChange: (limit: string) => void
+    }
+)
 
 export default function Pagination({
-  limit,
-  limits,
   page,
   hasNextPage,
   hasPreviousPage,
-  hideSelectors,
-  onLimitChange,
   onPageChange,
   ...props
 }: IProp) {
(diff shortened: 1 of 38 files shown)


components/Pagination/Pagination.tsx
```diff
@@ -8,187 +8,132 @@ import {
...
+} & (
+  | {
+      withoutLimit: true
+    }
+  | {
+      withoutLimit?: false
+      limit: number
+      limits: number[]
+      onLimitChange: (limit: string) => void
+    }
```
NicolasMahe, line 30: @ismailToyran Nice type definition ;)

components/Token/Owners/Modal.tsx
```diff
@@ -58,10 +57,20 @@ const OwnersModal: FC<Props> = ({
...
+  const hasNextPage = useMemo(
+    () => data?.ownerships?.pageInfo.hasNextPage,
+    [data?.ownerships?.pageInfo.hasNextPage],
+  )
+
+  const hasPreviousPage = useMemo(
+    () => data?.ownerships?.pageInfo.hasPreviousPage,
+    [data?.ownerships?.pageInfo.hasPreviousPage],
+  )
+
```
NicolasMahe, line 73: No need for a memo for those variable, no calculation is made inside the memo.
```suggestion
  const hasNextPage = data?.ownerships?.pageInfo.hasNextPage
  const hasPreviousPage = data?.ownerships?.pageInfo.hasPreviousPage
```
You could also simply remove use of `hasNextPage` / `hasPreviousPage` by directly `data?.ownerships?.pageInfo.hasNextPage` / `data?.ownerships?.pageInfo.hasPreviousPage`, they are used only once.

pages/users/[id]/bids/placed.tsx
```diff
@@ -77,16 +71,13 @@ const BidPlacedPage: NextPage<Props> = ({ now }) => {
     },
   })
 
-  const bidData = useMemo(() => data || previousData, [data, previousData])
-  const loading = _loading && !bidData
-
   const bids = useMemo(
     () =>
-      (bidData?.bids?.nodes || []).map((x) => ({
+      (data?.bids?.nodes || []).map((x) => ({
```
Comment (NicolasMahe) [inline pages/users/[id]/bids/placed.tsx:76]: The fallback to empty array could be removed so `bids` would also be undefined when `data?.bids?.nodes` is undefined.
```suggestion
      data?.bids?.nodes.map((x) => ({
```
This will allow to update the logic line 178 - 180 to use `bids` and be safer:
```tsx
          {bids === undefined ? (
            <Loader />
          ) : bids.nodes.length > 0 ? (
```

components/Pagination/Pagination.tsx
```diff
@@ -8,187 +8,132 @@ import {
...
-        {!isMobile && <Skeleton w={3 / 8} h={10} />}
-        <Skeleton w={isMobile ? 5 / 8 : 2 / 8} h={8} />
+      <Flex align={{ base: 'flex-end', sm: 'center' }} justify="space-between">
+        <Skeleton w={3 / 8} h={10} />
+        <Skeleton w={2 / 8} h={8} />
       </Flex>
     )
 
-  if (total === 0) return <></>
+  if (!hasNextPage && !hasPreviousPage) return null
```
NicolasMahe, line 50: There is an edge case because of this condition. The pagination is not displayed if there is no other page. This is weird when the user change the number of page from a multi-page query to a unique page:

https://github.com/liteflow-labs/starter-kit/assets/5823445/6b59992c-f239-4334-b2ae-1e55ba7806c4

I would let the display of pagination to be managed by the parent component page that already implement the empty state if that's **not too complicated**. The pagination should only NOT be display with the empty state.
If it's too complicated, let's keep it like this for now.

ismailToyran: [Change the way of displaying pagination](https://github.com/liteflow-labs/starter-kit/pull/367/commits/8ffc629ecde30f5aaf76e89431c6132eec1cbffb)

components/Token/Owners/Modal.tsx
ismailToyran: [remove not needed useMemo usages](https://github.com/liteflow-labs/starter-kit/pull/367/commits/7b667d97d135fbe297a9e2b0a67608eb4d4cda1d)

pages/users/[id]/bids/placed.tsx
Comment (ismailToyran) [reply pages/users/[id]/bids/placed.tsx]: [simplify array checks](https://github.com/liteflow-labs/starter-kit/pull/367/commits/f5700bc8ca546c8ee17e25ce2fcebf436e8de13c)
Because of this one, data check on useMemo wasn't triggering loading state, so had to also push this commit which makes sense.
[useMemo dependency change to show loader correctly](https://github.com/liteflow-labs/starter-kit/pull/367/commits/e84ff65447de557f8caf36f070a3b17eb3c31437)

PR #371, opened 2023-08-15
Remove previous data and simplify loading logic

NicolasMahe:
### Description

- Remove previous data because Apollo manage a local cache already and to simplify the codebase
- Replace param loading from queries by checking for undefined. This make the code a lot safer (check for null and undefined) and simplify the logic

### Checklist

- [x] Base branch of the PR is `dev`

PR #373, opened 2023-08-16
Collection header query split

ismailToyran:
### Project organization
- Related [trello card](https://trello.com/b/33afSjTr/product-backlog)
- Dependant on #371 

### Description
Splits collection detail page metrics from the actual page load, so page will be accessible before the metrics.

### DIFF
--- a/components/Collection/CollectionHeader.tsx
+++ b/components/Collection/CollectionHeader.tsx
@@ -30,103 +30,104 @@ import { chains } from '../../connectors'
 import { formatAddress } from '../../utils'
 
 type Props = {
-  collection: Partial<{
-    address: string
-    chainId: number
-    name: string | null
-    description: string | null
-    image: string | null
-    cover: string | null
-    twitter: string | null
-    discord: string | null
-    website: string | null
-    deployerAddress: string
-    deployer: {
-      address: string
-      name: string | null
-      username: string | null
-      verified: boolean
-    } | null
-    totalVolume: string
-    totalVolumeCurrencySymbol: string
-    floorPrice: string | null
-    floorPriceCurrencySymbol: string | null
-    totalOwners: number
-    supply: number
-  }> // TODO: add undefined type
-  loading: boolean // TODO: to delete
+  collection:
+    | {
+        address: string
+        chainId: number
+        name: string
+        description: string | null
+        image: string | null
+        cover: string | null
+        twitter: string | null
+        discord: string | null
+        website: string | null
+        deployer: {
+          address: string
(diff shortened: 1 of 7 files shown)


components/Collection/CollectionHeader.tsx
```diff
@@ -295,33 +285,34 @@ const CollectionHeader: FC<Props> = ({ collection, loading, reportEmail }) => {
...
+                  >
+                    {block.value}
+                  </Text>
+                  <Text
+                    variant="subtitle2"
+                    title={block?.name}
+                    isTruncated
+                    color="gray.500"
+                  >
+                    {block?.name}
```
NicolasMahe, line 313:
```suggestion
                    title={block.name}
                    isTruncated
                    color="gray.500"
                  >
                    {block.name}
```

pages/collection/[chainId]/[id].tsx
```diff
@@ -148,11 +166,15 @@ const CollectionPage: FC<Props> = ({ now }) => {
...
-      />
+      {!collectionDetails ? (
+        <CollectionHeaderSkeleton />
+      ) : (
+        <CollectionHeader
+          collection={collectionDetails}
+          metrics={collectionMetrics}
+          reportEmail={environment.REPORT_EMAIL}
+        />
+      )}
```
Comment (NicolasMahe) [inline pages/collection/[chainId]/[id].tsx:177]: This logic is weird.
Only `collectionDetails` is controlling the display of the loader, but not `collectionMetrics`. Either both or neither variables should display the loader here.

Moreover, `CollectionHeaderSkeleton` contains skeleton for the metrics part, but `CollectionHeader` also contains it. Not good to have duplication.

I suggest you to move all the loader logic in `CollectionHeader`, make `collection` undefinable, and remove the component `CollectionHeaderSkeleton`. `CollectionHeader`.

Another way would be to move out of the component `CollectionHeader` the metrics (display and loader) in a new component `CollectionHeaderMetrics`. This might be better that my first suggestion.

Comment (ismailToyran) [reply pages/collection/[chainId]/[id].tsx]: Your 2nd suggestion actually simplified process a lot.
Removed `undefined` from metrics that way as well.
Let me know what do you think
[Simply CollectionHeader by splitting into two](https://github.com/liteflow-labs/starter-kit/pull/373/commits/b92de659aab50c4d75cd8dfd73002b8645b79184)

NicolasMahe (approved):
Looking good in term of UI.
In term of logic, I would have have created new components for the skeletons but simply put them in the `CollectionHeader` and `CollectionMetrics` components.

PR #383, opened 2023-08-24
Async environment

antho1404:
Adding support for async environment so we can load environments from anywhere like a CMS, a database...

PR #384, opened 2023-08-24
Fetch environment from infra

antho1404:
Dependency https://github.com/liteflow-labs/starter-kit/pull/383 and waiting for the new release of the API

Get the environment from the infra and not hardcoded anymore so users can change the settings of the starter kit directly in the `/settings` page that will be live soon


colors.ts
```diff
@@ -0,0 +1,46 @@
...
+    .toString(16)
+    .padStart(2, '0')
+  const b = Math.round(b1 + (b2 - b1) * ratio)
+    .toString(16)
+    .padStart(2, '0')
+
+  return `#${r}${g}${b}`.toUpperCase()
+}
+
+export function generatePalette(color: string, extra = 0.05) {
```
antho1404, line 23: This generates a palette extremely close to the one we had, but I couldn't find any way to reproduce the current palette, so I believe it was a wrong palette

Screenshot 2023-09-24 at 18 02 51

components/Navbar/Navbar.tsx
```diff
@@ -122,13 +118,10 @@ const DrawerMenu: FC<{
...
-                src={logo?.path || '/logo.svg'}
-                alt="Logo"
-                width={logo?.width || 139}
-                height={logo?.height || 32}
-                w={logo?.width ? `${logo.width}px` : '139px'}
-                h={logo?.height ? `${logo.height}px` : '32px'}
+              <img
+                src={LOGO}
+                alt={META_COMPANY_NAME}
+                style={{ height: '32px' }}
```
antho1404, line 124: @ismailToyran do you see any way to use the `Image` from next? We cannot have the size of it as we only know the file URL. We can only fix the height but not the width

pages/_document.tsx
```diff
@@ -34,13 +34,22 @@ class MyDocument extends Document {
     return (
       <Html lang={this.props.locale || 'en'}>
         <Head>
+          <link
+            rel="shortcut icon"
+            href={
+              this.props.__NEXT_DATA__.props.pageProps.environment.FAVICON ||
+              '/favicon.ico'
+            }
+          />
```
antho1404, line 43: @ismailToyran is this the proper way to integrate the favicon?

antho1404:
All platforms have their settings up to date, so when the version with this change is merged, we should have the proper data.

pages/_document.tsx
ismailToyran: That's is true. We just need to make sure, the user will upload `.ico` file though.
```html
  <link rel="shortcut icon" href="/favicon.ico" />
```

components/Navbar/Navbar.tsx
ismailToyran: Can't we let them upload logo here? So they can put width as well, as it's required?
Or we have to modify navbar image to be absolute with fixed width and height, and need to shrink image to contain.
There is no other way I guess.

PR #388, opened 2023-08-28
Remove dynamic auction expiration check

NicolasMahe:
### Description

Remove dynamic auction expiration check to stop re-rendering the whole page tree every second.

I finally choose to simply removing the check every seconds rather than implementing a clever solution (check the expiration in a separate `useEffect` and trigger the `useMemo` only when needed) to keep it simple and because the chance of having the auction expiring during the creator of this auction is on the page is very low. Changing the page or a refresh will of course re-execute the check. @antho1404 do you agree?

### Checklist


antho1404 (approved):
This is not ideal and does break the current behavior. 

That said auctions are not used much and even tho the auction is updated based on the expiration, not the bids so with the current behavior we may have an UX issue where the seller accept the auction thinking it's 100 USDC where in fact there is a new bid in between for 103USDC for exemple.

So I'm good to have a full refresh required

PR #389, opened 2023-08-28
Add graphql-eslint

NicolasMahe:
### Description

Add graphql-eslint to detect issues in the gql files.

I had to create a new `.eslintrc` specially for this new linter as some Typescript rules requires some config even if they are not executed.

@antho1404 can you confirm this direction.

### Checklist


antho1404 (approved):
I'm not a big fan of the additional eslint file but that does help to clean up the graphql so I'm good with that

PR #391, opened 2023-08-31
fix issue with currency autoreload

antho1404:
Fix the issue where the currency sometimes refreshes.
This issue is a bit tricky to detect/reproduce.

Flow to reproduce:
- Go on a create offer page (or bid)
- Select a currency that is not the default one
- Wait until your computer switches to the next minute
- Update the price
- Here you notice the currency changes

This issue is related to the fact that dates are not in memo, resulting in some refresh that should not happen.


antho1404:
This PR has been reverted as merged to main by mistake.
Replaced by https://github.com/liteflow-labs/starter-kit/pull/393

PR #397, opened 2023-09-04
Seo updates

ismailToyran:
### Project organization
- Related [trello card](https://trello.com/c/NInxUgJD/1054-client-ticket-g2a-seo-issues)

### Description
Improve overall SEO of the platform

### Checklist

### DIFF
--- a/components/Head.tsx
+++ b/components/Head.tsx
@@ -1,4 +1,5 @@
 import NextHead from 'next/head'
+import { useRouter } from 'next/router'
 import { FC, PropsWithChildren } from 'react'
 import useEnvironment from '../hooks/useEnvironment'
 
@@ -15,9 +16,18 @@ const Head: FC<PropsWithChildren<Props>> = ({
   children,
 }) => {
   const { META_TITLE, META_DESCRIPTION, BASE_URL } = useEnvironment()
+  const { asPath } = useRouter()
   const shortDescription = (description || META_DESCRIPTION).substring(0, 160)
   const siteTitle =
     title !== META_TITLE ? `${title} - ${META_TITLE}` : META_TITLE
+
+  // remove query string and hash from asPath
+  const _pathSliceLength = Math.min.apply(Math, [
+    asPath.indexOf('?') > 0 ? asPath.indexOf('?') : asPath.length,
+    asPath.indexOf('#') > 0 ? asPath.indexOf('#') : asPath.length,
+  ])
+  const canonicalURL = BASE_URL + asPath.substring(0, _pathSliceLength)
+
   return (
     <NextHead>
       <title>{siteTitle}</title>
@@ -31,6 +41,7 @@ const Head: FC<PropsWithChildren<Props>> = ({
         name="twitter:image"
         content={image || `${BASE_URL}/twitter-card.jpg`}
       />
+      <link rel="canonical" href={canonicalURL} />
       {children}
     </NextHead>
   )
--- a/components/Head.tsx
+++ b/components/Head.tsx
@@ -14,24 +14,18 @@ const Head: FC<PropsWithChildren<Props>> = ({
   image,
   children,
 }) => {
-  const { META_TITLE, BASE_URL } = useEnvironment()
-  const shortDescription = description?.substring(0, 160)
+  const { META_TITLE, META_DESCRIPTION, BASE_URL } = useEnvironment()
+  const shortDescription = (description || META_DESCRIPTION).substring(0, 160)
+  const siteTitle =
+    title !== META_TITLE ? `${title} - ${META_TITLE}` : META_TITLE
   return (
     <NextHead>
-      <title>
-        {title && title !== META_TITLE
-          ? `${title} - ${META_TITLE}`
-          : META_TITLE}
-      </title>
-      <meta property="og:title" content={title} />
-      <meta name="twitter:title" content={title} />
-      {shortDescription && (
-        <>
-          <meta name="description" content={shortDescription} />
-          <meta property="og:description" content={shortDescription} />
-          <meta name="twitter:description" content={shortDescription} />
-        </>
-      )}
+      <title>{siteTitle}</title>
+      <meta property="og:title" content={siteTitle} />
+      <meta name="twitter:title" content={siteTitle} />
+      <meta name="description" content={shortDescription} />
+      <meta property="og:description" content={shortDescription} />
+      <meta name="twitter:description" content={shortDescription} />
       <meta property="og:image" content={image || `${BASE_URL}/og-image.jpg`} />
       <meta
         name="twitter:image"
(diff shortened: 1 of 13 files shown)


components/Head.tsx
```diff
@@ -24,11 +25,11 @@ const Head: FC<PropsWithChildren<Props>> = ({
       </title>
       <meta property="og:title" content={title} />
       <meta name="twitter:title" content={title} />
```
NicolasMahe, line 27: Shouldn't the `og:title` and `twitter:title` have the same logic as the main title?
Maybe the issue I was speaking about on G2A is related to this?

public/robots.txt
```diff
@@ -2,3 +2,5 @@ User-agent: *
 Crawl-Delay: 10
 Disallow: /_next/image
 Disallow: /api/
+Disallow: /tokens/[id]/bid
+Disallow: /tokens/[id]/offer
```
ismailToyran, line 6: How can I check, if this actually works or not?
Another option is adding `<meta name="robots" content="noindex,nofollow" />` to the pages that we want.
What should be the best way?

pages/checkout/[id].tsx
```diff
@@ -91,7 +93,7 @@ const CheckoutPage: NextPage<Props> = ({ now }) => {
   return (
     <SmallLayout>
       <Head
-        title={asset ? t('offers.checkout.meta.title', asset) : ''}
+        title={asset ? t('offers.checkout.meta.title', asset) : META_TITLE}
```
Comment (ismailToyran) [inline pages/checkout/[id].tsx:96]: I can change this prop to be `title: string | undefined` so that way no need to pass `META_TITLE` all the time. What do you think? If we do that, that means that title is required anyway (not optional) so we still force to use title on `<Head />` component.

components/Head.tsx
ismailToyran: Might be. Fixed that already 👍

pages/checkout/[id].tsx
Comment (antho1404) [reply pages/checkout/[id].tsx:96]: Or you could make the whole `title` optional  I'm good to fallback on the `META_TITLE` if we don't have a title set so we don't have to read the env everywhere

public/robots.txt
antho1404: I don't think the [id] works indeed, maybe a `*` would work better but otherwise your solution of adding it directly to the head is good

pages/checkout/[id].tsx
Comment (ismailToyran) [reply pages/checkout/[id].tsx]: [Change title to be optional on Head component](https://github.com/liteflow-labs/starter-kit/pull/397/commits/0928e35592efdbf11fa8223b647c75fa9ee03b8e)

public/robots.txt
ismailToyran: [Change the way of applying robots](https://github.com/liteflow-labs/starter-kit/pull/397/commits/1602908dcf35a64abdb3c66c73731a00e7a3f6fa)

components/Head.tsx
```diff
@@ -14,28 +15,32 @@ const Head: FC<PropsWithChildren<Props>> = ({
...
+  const { asPath } = useRouter()
+  const shortDescription = (description || META_DESCRIPTION).substring(0, 160)
+  const siteTitle = title ? `${title} - ${META_TITLE}` : META_TITLE
+
+  // remove query string and hash from asPath
+  const _pathSliceLength = Math.min.apply(Math, [
+    asPath.indexOf('?') > 0 ? asPath.indexOf('?') : asPath.length,
+    asPath.indexOf('#') > 0 ? asPath.indexOf('#') : asPath.length,
+  ])
+  const canonicalURL = BASE_URL + asPath.substring(0, _pathSliceLength)
```
antho1404, line 28: For information you could use substring from the end to the beginning like that

```ts
asPath
  .substring(asPath.indexOf('?'), 0)
  .substring(asPath.indexOf('#'), 0)
```

This basically removes everything after `?` and after `#` (included)

Also it would be great to put this in a memo

components/Head.tsx
```diff
@@ -14,28 +15,32 @@ const Head: FC<PropsWithChildren<Props>> = ({
...
-      {description && (
-        <>
-          <meta name="description" content={description} />
-          <meta property="og:description" content={description} />
-          <meta name="twitter:description" content={description} />
-        </>
-      )}
+      <title>{siteTitle}</title>
+      <meta property="og:title" content={siteTitle} />
+      <meta name="twitter:title" content={siteTitle} />
```
antho1404, line 34: We don't want to have the same title on Twitter.
The title for the website is needed for the tab to provide context, like the page and the website of this tab, but in Twitter or when we share, there is already the website, so only the page is needed. That allows to focus only on the important part and is what most website are doing

public/robots.txt
antho1404: 👍

antho1404 (commented):
Just one change to do but otherwise that's great.
I tested tho to share on twitter or slack and the image of the collections don't show up.

Could you make sure to test this and that it is working
Screenshot 2023-09-06 at 10 28 09

components/Head.tsx
ismailToyran: Reverted with [Change title for twitter and og](https://github.com/liteflow-labs/starter-kit/pull/397/commits/9fa86bd4b3b5675e0b12cabf85511c0260996975)

components/Head.tsx
ismailToyran: There still needs to be check if there is an actual usage of `?` or `#`
My actual implementation works, your suggestion needs more modification.
I've put it inside `useMemo` with [Wrap canonicalURL with useMemo](https://github.com/liteflow-labs/starter-kit/pull/397/commits/5abeea92fb03b581a09e0cea0e0a2ae3d0892cc3)

ismailToyran:
I've fixed the issue with collection image not appearing on collection detail page. It was working on token detail page so didn't realise the issue on collection. Almost identical implementation. Should be good to go.

antho1404 (approved):

pages/checkout/[id].tsx
```diff
@@ -91,14 +91,13 @@ const CheckoutPage: NextPage<Props> = ({ now }) => {
...
-                name: asset.name,
-                creator: asset.creator.name || asset.creator.address,
-              })
-            : undefined
+          asset &&
+          t('offers.checkout.meta.description', {
+            name: asset.name,
+            creator: asset.creator.name || asset.creator.address,
+          })
         }
```
Comment (NicolasMahe) [inline pages/checkout/[id].tsx:101]: Let's also remove this page from search result:
```
<meta name="robots" content="noindex,nofollow" />
```

public/robots.txt
NicolasMahe: I would recommend to add those pages in the robots.txt to prevent the robot to even fetch the page in the first place.
The meta header are return after the robot already fetch the page. This is good to prevent the page to show in the search engine results but the page is still fetch nontheless.

It seems possible to use `*` in the `Disallow` rule:
```
Disallow: /tokens/*/bid
Disallow: /tokens/*/offer
```

PR #399, opened 2023-09-06
Improve auction flow

antho1404:
When visualizing the bids in the profile, the bids from auctions are displayed like any bids and trigger errors when trying to accept them because the auction is still in progress.

This adds context to the bid line and prevent user from accepting the bid in their profile


pages/users/[id]/bids/placed.gql
```diff
@@ -1,14 +1,19 @@
 query FetchUserBidsPlaced(
   $limit: Int!
   $offset: Int!
-  $orderBy: [OfferOpenBuysOrderBy!]
+  $orderBy: [OffersOrderBy!]
   $address: Address!
 ) {
-  bids: offerOpenBuys(
+  bids: offers(
```
Comment (antho1404) [inline pages/users/[id]/bids/placed.gql:7]: Need to switch to the `offers` API and add all the proper filters as `offerOpenBuys` does not contain the auction information cc @NicolasMahe

antho1404:
@ismailToyran, could you double-check that this change doesn't create any side effect that I may have miss

convert.ts
```diff
@@ -595,19 +595,23 @@ export const convertBid = (
...
-    Pick<OfferOpenBuy, 'id' | 'availableQuantity' | 'expiredAt' | 'createdAt'>,
+    Pick<Offer, 'id' | 'availableQuantity' | 'expiredAt' | 'createdAt'> & {
+      auctionId?: string | null
+    },
 ): Required<ReturnType<typeof convertBid>> & {
   id: string
   availableQuantity: BigNumber
   expiredAt: Date | undefined
   createdAt: Date
+  auctionId?: string
```
ismailToyran, line 606:
```suggestion
  auctionId: string | undefined
```

pages/users/[id]/bids/received.tsx
```diff
@@ -239,28 +245,38 @@ const BidReceivedPage: NextPage<Props> = ({ now }) => {
...
-                        )}
+                        {ownerLoggedIn &&
+                          (item.auctionId ? (
+                            <Button
+                              as={Link}
+                              variant="outline"
+                              colorScheme="gray"
+                              href={`/tokens/${item.asset.id}`}
+                            >
+                              {t('user.bid-received.actions.view-nft')}
```
Comment (ismailToyran) [inline pages/users/[id]/bids/received.tsx:256]:
```suggestion
                              {t('user.bid-received.actions.view')}
```

ismailToyran (approved):
If the filter is working correctly (which seems to me)
```
    filter: {
      type: { equalTo: BUY }
      availableQuantity: { greaterThan: "0" }
      signature: { isNull: false }
      makerAddress: { equalTo: $address }
    }
```
Looking good to me 👍

convert.ts
antho1404: This is to ensure that other functions calling this work as they still use the `openOfferBuy` and not `offers` API, which does not have the `auctionId` so we won't be able to have the field here

pages/users/[id]/bids/placed.gql
Comment (NicolasMahe) [reply pages/users/[id]/bids/placed.gql:7]: Filters are correct 👍

PR #403, opened 2023-09-06
Replace refetch on page change of NavbarAccount query by polling

NicolasMahe:
### Description

Replaced `refetch` on page change of `NavbarAccount` query by polling.
The `refetch` function bypass the `skip` parameter resulting in this query to be executed even when user is not connected and change the page.
I replace it by a polling of 60sec because re-executing this query is only for updating the notification icon. Login will trigger this query anyway.

@antho1404 what do you think?

It would still be possible to keep the `refetch` on page change but need to manually check the `skip` param to not call `refetch` when user is not connected, but that will result is a still high number of queries for not so much benefit.

### Checklist

### DIFF
--- a/components/Navbar/Navbar.tsx
+++ b/components/Navbar/Navbar.tsx
@@ -342,21 +342,21 @@ const Navbar: FC<{
   const { t } = useTranslation('components')
   const { address, isLoggedIn, logout, isConnected } = useAccount()
   const { disconnect } = useDisconnect()
-  const { asPath, query, push, isReady, events } = useRouter()
+  const { asPath, query, push, isReady } = useRouter()
   const { register, setValue, handleSubmit } = useForm<FormData>()
   const [cookies] = useCookies()
   const { openConnectModal } = useConnectModal()
   const lastNotification = cookies[`lastNotification-${address}`]
   const {
     data: accountData,
-    refetch,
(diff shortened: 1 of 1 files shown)


antho1404:
I would suggest to revert this PR, this creates an issue. The point of having this called when a page changes was to fix a previous issue when a profile was updated, the navbar wasn't updated. So now if we update the profile, the image is not updated and it looks like the edition of the profile didn't work. I would suggest to actually do a proper fix to only call the refetch on page change if there is an account

NicolasMahe:
> So now if we update the profile, the image is not updated and it looks like the edition of the profile didn't work

The image is updated after the update of profile, I don't see any issue on my side. Do you have an issue when updating? This is done by the query `FetchAccountDetail` triggered when the user is redirected to its profile page. This query update the local cache corresponding to the account and Apollo trigger a re-render of the navbar.

---

I do see a sustain number of query, I guess from logged in users with website open in a tab in background:

Screenshot 2023-09-08 at 09 39 16

The best would be to implement focus detection with `document.hasFocus()` and some kind of debounce to refetch after the browser get the focus again but not less than the poll duration.
Because it looks more complex that the refetch on page change, and the original issue was just to not trigger this query when user is not logged in, I will put back the refetch on page change only when the user is logged in.

PR #405, opened 2023-09-06
Fetch owners with LazyQuery

ismailToyran:
### Project organization
- Part of [trello card](https://trello.com/c/POSj276d/1028-investigate-high-api-and-service-calls)

### Description
Only fetch owners when user click on the modal button

### Checklist


components/Token/Owners/ModalActivator.tsx
```diff
@@ -54,7 +54,7 @@ const OwnersModalActivator: FC<Props> = ({
             bgColor="brand.50"
           >
             <Text as="span" variant="caption" color="brand.500">
-              {`+${numberOfOwners >= 103 ? 99 : owners.length - 4}`}
+              {`+${numberOfOwners >= 103 ? 99 : numberOfOwners - 4}`}
```
ismailToyran, line 57: Small fix with the number, as we limit owners to 5, this is no longer showing the true count.

components/Token/Owners/Modal.tsx
```diff
@@ -58,6 +59,13 @@ const OwnersModal: FC<Props> = ({
     },
   })
 
+  const openOwners = useCallback(async () => {
+    onOpen()
+    if (!called) {
```
NicolasMahe, line 64: Do we need this check on called here? I'm a bit worry that this will introduce side effect somehow.

NicolasMahe (approved):
Working great but I put one comment about a possible side effect.

PR #406, opened 2023-09-07
Merge queries `FetchAssetBids` and `FetchAuctionBids` into `FetchAsset`

NicolasMahe:
### Description

Merge queries `FetchAssetBids` and `FetchAuctionBids` into `FetchAsset` as one of them (mostly `FetchAssetBids`) is trigger every time the page token detail is rendered.

I keep the `FetchAssetHistory` separated because it's only executed when the user select the `history` tab and it may be slow if lots of transfers are related to the token. It will be easier to track its performance and use if separated.

### Checklist


pages/tokens/[id]/index.gql
```diff
@@ -53,12 +53,27 @@ query FetchAsset(
         status
       }
     }
-    bestBid: bids(
```
Comment (NicolasMahe) [inline pages/tokens/[id]/index.gql:56]: The `bestBid` was never used by type was required by function `convertTraits` which didn't use it as well.

PR #419, opened 2023-09-20
New SearchInput added

ismailToyran:
### Project organization
- Related [trello card](https://trello.com/c/WXYX8kgK/1068-client-ticket-g2a-page-not-refreshed-after-clear-search-input)

### Description
- Remove default behaviour of browsers clear input action
- Add custom search input to add new behaviour mentioned above
- Now, when you click on clear icon on search bar, it will reset search results instead of just clearing input.

### Checklist

### DIFF
--- a/components/SearchInput.tsx
+++ b/components/SearchInput.tsx
@@ -3,16 +3,23 @@ import { HiOutlineSearch } from '@react-icons/all-files/hi/HiOutlineSearch'
 import { HiOutlineX } from '@react-icons/all-files/hi/HiOutlineX'
 import { useRouter } from 'next/router'
 import { FC, HTMLAttributes, useCallback, useMemo } from 'react'
-import { useFormContext } from 'react-hook-form'
+import { RegisterOptions, useFormContext } from 'react-hook-form'
 import { removeEmptyFromObject } from '../utils'
 
 type IProps = HTMLAttributes<any> & {
   name: string
+  registerOptions?: RegisterOptions
   onReset?: () => void
   onSubmit?: () => void
 }
 
-const SearchInput: FC<IProps> = ({ name, onReset, onSubmit, ...props }) => {
+const SearchInput: FC<IProps> = ({
+  name,
+  registerOptions,
+  onReset,
+  onSubmit,
+  ...props
+}) => {
   const { query, push, pathname } = useRouter()
   const { register, watch } = useFormContext()
   const search = watch(name)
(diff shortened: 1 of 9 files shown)


ismailToyran:
Hi Anthony, we've different usages of search input, so I had to tweak a bit.
What do you think about the current implementation?
- Removed 2 `setState` to use `react-hook-form` instead. Had to wrap with `<FormProvider />` (to be honest, I think we should always do that, so the children could also see related form and it's values)
- The only downside (maybe?) is now, I've to exclude those from filter so we don't end up something like `collectionSearch=&propertySearch=` after reset
- Maybe for the consistency I should keep `{...register()}` but `name` prop was needed for some checks.

I've tested every flow I could think of, and it seems to be working as intended without any issue.

components/SearchInput.tsx
```diff
@@ -0,0 +1,60 @@
...
+      page: undefined,
+    })
+    await push({ pathname, query: cleanData }, undefined, {
+      shallow: true,
+    })
+  }, [onReset, push, query, pathname])
+
+  return (
+    <InputGroup>
+      <Input {...register(name)} type="search" pr={14} {...props} />
```
antho1404, line 35:
```suggestion
      <Input type="search" name={name} pr={14} {...props} />
```

components/SearchInput.tsx
```diff
@@ -0,0 +1,60 @@
+import { Icon, Input, InputGroup, InputRightElement } from '@chakra-ui/react'
+import { HiOutlineSearch } from '@react-icons/all-files/hi/HiOutlineSearch'
+import { HiOutlineX } from '@react-icons/all-files/hi/HiOutlineX'
+import { useRouter } from 'next/router'
+import { FC, HTMLAttributes, useCallback, useMemo } from 'react'
+import { useFormContext } from 'react-hook-form'
+import { removeEmptyFromObject } from '../utils'
+
+type IProps = HTMLAttributes<any> & {
```
antho1404, line 9: We could add directly the results of the `register` function in props so we could let the parent component do the `...register` as you describe for consistency and also to remove some logic from this component

```suggestion
type IProps = HTMLAttributes<any> & ReactFormRegisterReturnData {
```

antho1404 (approved):
Working well, just one feedback but I let you take the decision for that.
I think it would be nicer to have the `register` in the parent component and have something like `<SearchInput {...register('search')} />` we should always be able to get the name from the props as the `register` function does return the name and this way we have a similar codebase everywhere and might also prevent some issue in the future if we want to pass more props to some search or do some validation for exemple.

Other than that good job, that's working well and that's you call if you want to merge or do this change ;)

components/SearchInput.tsx
ismailToyran: I've tried so many different usages. But unfortunately none of them works.
You will probably remember about wrapping component with `Controller` from the past. Which I don't want to go in that direction.
Also your suggestion + `useRef` didn't work either.
This is the only way I was able to make it work actually.

Even though that's not what we want in the end, I've added optional `registerOptions` in case we need to add some validation [Add registerOptions](https://github.com/liteflow-labs/starter-kit/pull/419/commits/eca1408a9b68eb3452abe8d2e10986b21a9ebada)

So I will move forward with this one. But I'm pretty sure this is the correct way of registering the input (if we intend to modify and add functionality to use it as a new Component)

PR #421, opened 2023-09-20
fix Unknown event handler property  error

antho1404:
Because the `props` was not fully destructured, somehow react couldn't detect the props and we had the following error
This PR removes the conditional type that leads to the `props` being not fully destructured, and this solves the issue

> Warning: Unknown event handler property `onLimitChanged`. It will be ignored

@ismailToyran let me know if you have any other way to fix that better while preserving the type, I don't really like the optional part but that's ok


components/Pagination/Pagination.tsx
```diff
@@ -19,23 +19,21 @@ export type IProp = {
...
-      withoutLimit?: false
-      limit: number
-      limits: number[]
-      onLimitChange: (limit: string) => void
-    }
-)
+  onLimitChange?: (limit: string) => void
+  withoutLimit?: boolean
+  limit?: number
+  limits?: number[]
```
ismailToyran, line 25: My only suggestion, I think, will be more complicated.
```
  limit?: {
    base: number
    limits: number[]
    onLimitChange: (limit: string) => void
  }
```
Was thinking of something like this, but let's keep what you've changed

PR #431, opened 2023-09-28
Add mimetype detection of token media

NicolasMahe:
### Project organization

- Closes 
- Related 
- Dependant on 

### Description

Re-organize the logic of which media to display and implement a proper mimetype detection using a HEAD http query to access the `Content-Type` header.

The suggested fix to simply detect a `.html` extension in the media URL is not enough as most of the tokens of the collection with the issue are IPFS url without any `.html` extension, eg: https://arcade-genesis.b-cdn.net/QmcoSoBwimD7bCLMRYWEJAWAuiHe3DSMyqZ8aMmSEycWYj/500.json

- https://github.com/liteflow-labs/starter-kit/commit/933157fb6fdaf6a9463a25aba41fb116892694df: Define in advance the ordered list of media to display with their estimated mimetype. On error, try the next item in the list of media to disaply.
- https://github.com/liteflow-labs/starter-kit/commit/3ced3301b25cac5d945d22ecfd0ab297d9acf8e6: Add mimetype detection of token media using an HEAD http query. Fallback on error to hardcoded mimetype to still try to display the media.

### Checklist

### DIFF
--- a/components/Token/Media.tsx
+++ b/components/Token/Media.tsx
@@ -3,6 +3,7 @@ import {
   Center,
   chakra,
   Icon,
+  Skeleton,
   Stack,
   Text,
   useTheme,
@@ -10,6 +11,8 @@ import {
 import { FaImage } from '@react-icons/all-files/fa/FaImage'
 import { FC, useCallback, useEffect, useMemo, useState } from 'react'
 import Image from '../Image/Image'
+
+const authorizedMimetypes = ['image/', 'video/']
 
 const TokenMedia: FC<{
   imageUrl: string
@@ -31,7 +34,9 @@ const TokenMedia: FC<{
   const { colors } = useTheme()
 
   // setup content to display in order with mimetype. The goal is if there is an issue with one, try the next one until no more
-  const contentsToDisplay = useMemo<{ url: string; mimetype: string }[]>(() => {
+  const contentsToDisplay = useMemo<
+    { url: string; mimetype?: string; fallbackMimetype?: string }[]
+  >(() => {
     const contentsToDisplay = []
     if (unlockedContent) {
       if (unlockedContent.mimetype) {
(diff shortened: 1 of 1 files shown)


components/Token/Media.tsx
```diff
@@ -55,25 +33,131 @@ const TokenMedia: FC<{
...
 
-  const onImageError = useCallback(() => setImageError(true), [])
-  const onVideoError = useCallback(() => setVideoError(true), [])
+    // fetch mime type
+    fetch(toDisplay.url, {
+      method: 'HEAD',
+      // TODO: add request cancelation
+    })
+      .then((response) => {
+        const mimetype = response.headers.get('Content-Type')
```
NicolasMahe, line 133: Should also check the lowercased version of the header:
```suggestion
        const mimetype = response.headers.get('Content-Type') || response.headers.get('content-type')
```

components/Token/Media.tsx
```diff
@@ -55,25 +33,131 @@ const TokenMedia: FC<{
...
+      // do not try to fetch mime type if url is not http or https
+      return
+    }
 
-  const onImageError = useCallback(() => setImageError(true), [])
-  const onVideoError = useCallback(() => setVideoError(true), [])
+    // fetch mime type
+    fetch(toDisplay.url, {
+      method: 'HEAD',
+      // TODO: add request cancelation
```
NicolasMahe, line 130: Important todo here

components/Token/Media.tsx
```diff
@@ -55,25 +33,131 @@ const TokenMedia: FC<{
...
+      })
+    }
+    contentsToDisplay.push({
+      url: imageUrl,
+      fallbackMimetype: 'image/*',
+    })
+    contentsToDisplay.push({
+      url: imageUrl,
+      fallbackMimetype: 'video/*',
+    })
```
NicolasMahe, line 77: The double rows with a different `fallbackMimetype` is not the best with the mimetype detection system as it may execute two times the http HEAD request if the mimetype fetched is not authorized (it will call `onError` function that will trigger the http HEAD request with the next row of `contentsToDisplay`).
The solution could be to change the structure of data of `contentsToDisplay` to put an array of `fallbackMimetype` and update the display/retry logic. Could also split this component into a sub-component that would be responsible for fetching and trying the mimetypes, the main component would be responsible for the defining the order of the media to display and switch from one media to the next if the media cannot be displayed.

NicolasMahe:
Indeed there is a still a CORS on some of the NFTs of this collection. I must have miss those ones. This CORS protection is going to prevents the detection of mimetype and throw even more error in the console making the system pretty much worst.

Then I don't see any way to fix this issue client-side, we have to do it backend-side, save the mimetype in database, resync all NFTs, and expose it on the API. That's not going to be a quick fix, specially the resync of NFTs part.

PR #435, opened 2023-09-29
Improve token media component with fallback and better detection

antho1404:
Improve token component to better handle the different cases and file types
The `TokenMedia` component now only accepts a `media` and an optional `fallback`.
This is the responsibility of the parent component to pass the proper media, which can be retrieved by the new `useDetectAssetMedia` hook
The component also automatically does the fallback if the media fails.

You can check the following tokens on uaan
- Simple image: 97-0xcfbb45ff528c8ff6542887f631a1f704b92cf7db-34081265737503256337038727118962159868999502952850231594707337758359078404481
- Video with preview: 97-0xcfbb45ff528c8ff6542887f631a1f704b92cf7db-34081265737503256337038727118962159868999502952850231594724415393416730724742
- Private image (feel free to purchase it to unlock): 97-0xcfbb45ff528c8ff6542887f631a1f704b92cf7db-34081265737503256337038727118962159868999502952850231594729382497880446047239
- Private video (feel free to purchase it to unlock): 97-0xcfbb45ff528c8ff6542887f631a1f704b92cf7db-34081265737503256337038727118962159868999502952850231594746655469053395110262
- html file as image (not supported): 97-0xcfbb45ff528c8ff6542887f631a1f704b92cf7db-34081265737503256337038727118962159868999502952850231595345751279127311432018
- html file as animation with fallback to image: 97-0xcfbb45ff528c8ff6542887f631a1f704b92cf7db-34081265737503256337038727118962159868999502952850231595372123532009324221273
- html file as animation with fallback to video: 97-0xcfbb45ff528c8ff6542887f631a1f704b92cf7db-34081265737503256337038727118962159868999502952850231599801639647930375296097
- html file as animation with fallback to html file (not supported): 97-0xcfbb45ff528c8ff6542887f631a1f704b92cf7db-34081265737503256337038727118962159868999502952850231599814605654957745054776

I also explore some other ways https://github.com/liteflow-labs/starter-kit/pull/434 but that will require more work to make it more efficient but the UX will be much nicer


hooks/useDetectAssetMedia.ts
```diff
@@ -0,0 +1,50 @@
...
+      }
+    }
+    if (asset.unlockedContent) return asset.unlockedContent
+    if (asset.animationUrl) return { url: asset.animationUrl }
+    return { url: asset.image }
+  }, [asset])
+
+  const fallback = useMemo(() => {
+    if (!asset) return null
+    if (asset.animationUrl) return { url: asset.image }
```
antho1404, line 42:
```suggestion
    if (asset.image) return asset.image
```

hooks/useDetectAssetMedia.ts
```diff
@@ -0,0 +1,50 @@
...
+      }
+    }
+    if (asset.unlockedContent) return asset.unlockedContent
+    if (asset.animationUrl) return { url: asset.animationUrl }
+    return { url: asset.image }
+  }, [asset])
+
+  const fallback = useMemo(() => {
+    if (!asset) return null
+    if (asset.image) return asset.image
```
antho1404, line 42:
```suggestion
    if (asset.animation) return { url: asset.image } 
```

NicolasMahe (approved):
Working good 👍 No rendering issue with the different media types.
I put two suggestions, feel free to implement or discard them.

hooks/useDetectAssetMedia.ts
```diff
@@ -0,0 +1,50 @@
...
+    if (asset.unlockedContent) return asset.unlockedContent
+    if (asset.animationUrl) return { url: asset.animationUrl }
+    return { url: asset.image }
+  }, [asset])
+
+  const fallback = useMemo(() => {
+    if (!asset) return null
+    if (asset.animationUrl) return { url: asset.image }
+    return null
+  }, [asset])
```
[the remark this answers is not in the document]
antho1404: Yes I was thinking about that but I find it weird to fallback on the public asset. 
If there is an issue, in that case I believe the user should see the issue because they might have purchased this asset for the unlockable content

hooks/useDetectAssetMedia.ts
```diff
@@ -0,0 +1,50 @@
+import { useMemo } from 'react'
+
+export type FileResult = {
```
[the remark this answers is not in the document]
antho1404: Done

components/Token/Media.tsx
```diff
@@ -126,27 +76,16 @@ const TokenMedia: FC<{
     )
   }
 
-  // Use a basic image when the file is a blob or data
-  if (image.startsWith('blob:') || image.startsWith('data:'))
```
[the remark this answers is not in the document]
antho1404: Yes this was an issue in an "old" version of next that was fixed later, and we did update the next version without removing this "hack" that we had. That was not needed for a long time already

components/Drop/DropCard.tsx
```diff
@@ -63,10 +63,12 @@ export default function DropCard({ drop, onCountdownEnd }: Props) {
...
+              url: drop.collection.cover,
+              mimetype: null,
+            }}
+            fallback={null}
             defaultText={drop.collection.name}
-            animationUrl={undefined}
-            unlockedContent={null}
             sizes="(min-width: 62em) 600px, 100vw"
             fill
           />
```
antho1404, line 74: I don't think the drop should use the `TokenMedia` as token media is for a token, I think a normal `Image` should do the job similar to the collection or user page

antho1404: moved here https://github.com/liteflow-labs/starter-kit/issues/440

PR #437, opened 2023-09-29
Bypass not supported medias

antho1404:
Dependency https://github.com/liteflow-labs/starter-kit/pull/435

Get mimetype from API and automatically fallbacks if the media is not supported


package.json
```diff
@@ -60,7 +60,7 @@
     "@graphql-codegen/typescript-react-apollo": "^3.3.7",
     "@graphql-eslint/eslint-plugin": "^3.20.1",
     "@next/bundle-analyzer": "^13.4.9",
-    "@nft/api-graphql": "^1.0.0-beta.34",
+    "@nft/api-graphql": "^1.0.0-beta.41-prerelease-4",
```
antho1404, line 63: @NicolasMahe please update the package when released

antho1404 (commented):
I'm not sure everything has been deployed, I tested with the different assets in https://github.com/liteflow-labs/starter-kit/pull/435 and I still have the CORB error because we don't have the mimetype (which is normal) but I refreshed the assets and still the mimetype is `null`

@NicolasMahe could you deploy everything or check what's wrong.

components/Token/Media.tsx
```diff
@@ -29,14 +34,21 @@ const TokenMedia: FC<{
...
   }, [imageError, videoError, fallback, mediaToDisplay])
 
+  // Switch to the fallback when the media is not supported
+  useEffect(() => {
+    const mimetype = mediaToDisplay?.mimetype
+    if (!mimetype) return // assume it's supported
+    if (supportedMedia.some((regex) => regex.test(mimetype))) return // it's supported
+    if (mediaToDisplay?.url === fallback?.url) return
+    setMediaToDisplay(fallback)
+  }, [fallback, mediaToDisplay])
```
NicolasMahe, line 44: This doesn't prevent media with not supported mimetype to be "tried" to be display and throwing error in the console. There is a request to the html media before this `useEffect` is executed.

You can check on the following token: https://nft-test-uaan-git-feature-mimetype-detection-liteflow.vercel.app/tokens/1-0xa0c38108bbb0f5f2fb46a2019d7314cce38f3f22-10 (that's the collection with html file issues).

I believe adding a check on the mimetype line 91 will be required:
```ts
  if (
    !mediaToDisplay.mimetype ||
    mediaToDisplay.mimetype.startsWith('image/')
  ) {
    return (
      <Box position="relative" w="full" h="full">
        <Image
          src={mediaToDisplay.url}
          alt={defaultText}
          onError={onImageError}
          fill
          objectFit={fill ? 'cover' : 'contain'}
          sizes={sizes}
          unoptimized={mediaToDisplay.mimetype === 'image/gif'}
        />
      </Box>
    )
  }
```

But then, no media is display after the initial loading until the `useEffect` is executed, that's why I added a loader in PR 431 to "mimic" the initial loader when the asset is loading:
https://github.com/liteflow-labs/starter-kit/pull/431/files#diff-6826b9eab3397084e54d012dd4f022c3e72bddaa990d74d2c491c79645ad04f1R225-R226

---

Another way would be to replace the two `useEffect` that call `setMediaToDisplay` by a better `useMemo` that will return the correct `mediaToDisplay` by checking for errors and mimetype.

NicolasMahe (changes requested):

components/Token/Media.tsx
antho1404: I like the idea of a memo, that was my initial implementation, but it makes everything complicated with a single component. Ideally, we should have a component handling the media/fallback and another for the video/image, and that should be possible, but with the current structure, that won't be possible with a memo.

Good catch on the first render (even tho the request is canceled). Adding the test as you describe will most likely create issues with a token like this https://nft-test-uaan-git-feature-mimetype-detection-liteflow.vercel.app/tokens/97-0xcfbb45ff528c8ff6542887f631a1f704b92cf7db-34081265737503256337038727118962159868999502952850231595345751279127311432018 that will load indefinitely if the media is correctly detected. Which point to my main question you haven't answered on the PR about the media detection that does not work

components/Token/Media.tsx
```diff
@@ -4,6 +4,8 @@ import { FC, useCallback, useEffect, useState } from 'react'
 import { AssetMedia } from '../../hooks/useDetectAssetMedia'
```
NicolasMahe, line 4: Creating a new thread to answer the following comment:
> I'm not sure everything has been deployed, I tested with the different assets in https://github.com/liteflow-labs/starter-kit/pull/435 and I still have the CORB error because we don't have the mimetype (which is normal) but I refreshed the assets and still the mimetype is null

> @NicolasMahe could you deploy everything or check what's wrong.

The refresh of asset is bypassed on lazy mint asset that have not been minted on chain (not purchased yet). I will check why we implement it like this and try to call syncAsset job anyway. I also need to check how the lazy mint asset are created, I may have forgot to set the mimetype at that time.

In the meantime, you can a non lazy mint asset or purchase the lazy mint asset and then refresh it.

antho1404: Thanks indeed, I didn't realize that lazyminted was different, I will mint them so I can finalize the testing for all cases

package.json
NicolasMahe: done

components/Token/Media.tsx
antho1404: I was able to test everything by minting all the assets, no issue with mimetype detections after mint

components/Token/Media.tsx
antho1404: In fact, because the format is tested before, that actually displays a normal error, so all good. We had a problem tho with infinite loading if the fallback was also an unsupported file like 97-0xcfbb45ff528c8ff6542887f631a1f704b92cf7db-34081265737503256337038727118962159868999502952850231599814605654957745054776 but that has been fixed

NicolasMahe (approved):
Working good 👍 

I would love to spend more time to replace the `useEffect` logic with a unique `useMemo` to avoid the extra loading state (skeleton) when the media's mimetype is known but incompatible (can see this with the SSR of https://nft-test-uaan-git-feature-mimetype-detection-liteflow.vercel.app/tokens/1-0xa0c38108bbb0f5f2fb46a2019d7314cce38f3f22-10) but I cannot find an simple way because of the `imageError` and `videoError` logic.

The current result is already very good and good to merge 👍

PR #359, opened 2023-08-09
Feature / Drops

ismailToyran:
### Project organization
- Related [trello card](https://trello.com/c/FFkLOSNr/947-growth-hacking-drops)

### Description
Adds drop feature to the starter kit. You can navigate to the drops page to see upcoming, in progress and ended drops.
And drop detail page by clicking cards on drops page to navigate and participate on the drop.

### Checklist
- [x] Base branch of the PR is `dev`
- [x] Add locales
- [x] Add skeleton
- [x] Finalize and merge https://github.com/liteflow-labs/liteflow-js/pull/175 to use it from liteflow package

### Todo
- [x] Need to release new version of liteflow-js. Build will pass when we release and update `@liteflow/core` and `@liteflow/react` packages

PR #438, opened 2023-09-30
Drops improvements

antho1404:
Improvements on the code for the feedback provided in https://github.com/liteflow-labs/starter-kit/pull/359


components/Drop/ListItem.tsx
```diff
@@ -39,13 +39,13 @@ type Props = {
...
 const DropListItem: FC<Props> = ({ drop, isOpen }) => {
   const { t } = useTranslation('components')
-  const timeline = useDropTimeline({ drop })
+  const state = useDropStatus(drop)
 
-  const timelineText = useMemo(() => {
-    if (timeline === Timeline.UPCOMING) return t('drop.timeline.upcoming')
-    if (timeline === Timeline.INPROGRESS) return t('drop.timeline.in-progress')
+  const stateText = useMemo(() => {
+    if (state === Status.UPCOMING) return t('drop.state.upcoming')
```
ismailToyran, line 45:
```suggestion
    if (state === Status.UPCOMING) return t('drop.timeline.upcoming')
```

pages/drops.tsx
```diff
@@ -76,26 +77,17 @@ const DropsPage: NextPage<Props> = ({ now }) => {
...
               columns={{ base: 1, md: 2 }}
               spacing={3}
               w="full"
               mb={endedDrops.length > 0 ? 8 : 0}
             >
-              {inprogressDrops.map((drop) => (
-                <DropCard
-                  key={drop.id}
-                  drop={drop}
-                  timeline={Timeline.INPROGRESS}
```
ismailToyran, line 90: This is the only one I would like to keep as it's on previous PR.

As adding `const status = useTimeStatus(drop)` to the `<DropCard />` component, this will trigger each second.
As we already know `status`, let's just pass that and avoid using `const status = useTimeStatus(drop)` for `<DropCard />`? 

Yes it looks better codewise, but performance wise not? What do you think @antho1404 ?

antho1404: I don't think that would be a massive issue, but I'm good to revert back to the initial code with hardcoded value

ismailToyran: Keeping that as it's 👍

PR #447, opened 2023-10-08
Removal of extra prop passing (WIP)

ismailToyran:
### Description
Remove extra prop passing as we no longer use Storybook.
Simplify prop passing

### Checklist


components/Referral/Form.tsx
```diff
@@ -1,74 +0,0 @@
-import { Button, Icon, Text, useToast } from '@chakra-ui/react'
```
ismailToyran, line 1: Not used component

components/Token/Supply.tsx
```diff
@@ -7,24 +7,17 @@ import { FC } from 'react'
 const Supply: FC<{
   current: BigNumber
   total: BigNumber
-  small?: boolean
```
ismailToyran, line 10: Not used prop

components/Wallet/Address.tsx
```diff
@@ -34,19 +33,6 @@ const WalletAddress: FC<{
     [address, formatted, isShort],
   )
 
-  const CustomTag: any = as
```
ismailToyran, line 37: Not used

pages/tokens/[id]/index.tsx
```diff
@@ -155,44 +138,9 @@ const DetailPage: NextPage<Props> = ({ now: nowProp }) => {
...
-      ...convertAuctionFull(first),
-      bids: first.offers.nodes.map(convertBidFull),
-    }
-    if (!auction) return
-    // check if auction is expired
-    if (new Date(auction.expireAt) <= new Date()) return
-    // check if auction has a winning offer
-    if (!!auction.winningOffer?.id) return
-    return auction
-  }, [asset])
```
Comment (ismailToyran) [inline pages/tokens/[id]/index.tsx:171]: This is the only logic change on this PR,
Instead of this complex logic, using hook to check if it's valid or not (where we already use this hook on other components already)

Removed this line of code and now just using `auction && isValid` check

ismailToyran:
I've checked every changes twice Anthony. I'm pretty happy with the results.
With these changes prop passing is minimal, still could be improved, but should be enough for now.
Pages with token cards, user cards etc. All look pretty good right now.

I've put comment for every single addition or deletion. Rest is only prop passing.
Component files might take a bit time to review. But pages should be pretty fast.

Sorry for kinda big PR review 😅

convert.ts
```diff
@@ -1,936 +0,0 @@
-// Temporary file to centralize all the conversions.
-// This file will be removed when the whole migration of the components is finalized
```
antho1404, line 2: 🎉 
I just see we had a comment for that, it was a long temporary file haha, glad that it's gone

pages/drops.tsx
```diff
@@ -30,25 +30,10 @@ const DropsPage: NextPage<Props> = ({ now }) => {
...
-    [inprogressDrops, upcomingDrops],
-  )
-
-  const endedDrops = useMemo(
-    () => data?.ended?.nodes.map((drop) => convertDropEnded(drop)) || [],
-    [data?.ended?.nodes],
-  )
+  const inprogressDrops = data?.inProgress?.nodes || []
+  const upcomingDrops = data?.upcoming?.nodes || []
+  const activeDrops = [...inprogressDrops, ...upcomingDrops]
```
antho1404, line 35: This one could be great to memoize as the array will always generate a new object thus rerendering all the time

pages/tokens/[id]/index.tsx
```diff
@@ -102,48 +93,31 @@ const DetailPage: NextPage<Props> = ({ now: nowProp }) => {
...
-      image: { url: data.asset.image, mimetype: data.asset.imageMimetype },
-      animation: data.asset.animationUrl
-        ? {
-            url: data.asset.animationUrl,
-            mimetype: data.asset.animationMimetype,
-          }
-        : null,
-    }
+    return data.asset
   }, [data])
```
Comment (antho1404) [inline pages/tokens/[id]/index.tsx:97]: Could simply be removed

pages/tokens/[id]/index.tsx
Comment (antho1404) [reply pages/tokens/[id]/index.tsx:171]: I'm good with that, I'm not sure this is exactly the same behavior but I did a few tests and couldn't see any issues so all good with that and it makes a lot more sense

antho1404 (approved):
Really good work, a lot of changes, definitely a long review but I think worse it, the codebase is so much cleaner now

pages/tokens/[id]/index.tsx
Comment (ismailToyran) [reply pages/tokens/[id]/index.tsx]: [Remove useMemo and simplify asset](https://github.com/liteflow-labs/starter-kit/pull/447/commits/605a1710e6d753a762f35f3779ac2a8a9653a583)

components/Drop/DropCard.tsx
```diff
@@ -25,13 +27,45 @@ import Price from '../Price/Price'
...
+  drops: {
+    startDate: Date
+    endDate: Date
+    unitPrice: string
+    currency: {
+      decimals: number
+      symbol: string
+      image: string
+    }
+  }[]
```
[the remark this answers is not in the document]
ismailToyran: [Accept drop instead of drops for drop card](https://github.com/liteflow-labs/starter-kit/pull/447/commits/4cc3f1f72f28b1878f1a8660b946986dd7cc9184)

pages/drops.tsx
ismailToyran: [Move drops inside of useMemo](https://github.com/liteflow-labs/starter-kit/pull/447/commits/13de7ab5246ffa63c12a60001f6f2126f963456d)

components/Filter/FilterBy/Collection.tsx
```diff
@@ -60,7 +60,8 @@ const FilterByCollection: FC<Props> = ({
   const collections = collectionData?.collections?.nodes
 
   const collection = useMemo(() => {
-    if (selectedCollection) return selectedCollection
+    if (selectedCollection)
+      return selectedCollection as CollectionCardProps['collection']
```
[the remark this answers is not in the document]
ismailToyran: [Remove forced type definition](https://github.com/liteflow-labs/starter-kit/pull/447/commits/bae7daa03391e69a6b77f9c1a1501e4a631649a5)
Maybe not the most efficient for query but by passing correct props no need to hack typescript.

components/Collection/CollectionCard.tsx
```diff
@@ -9,20 +9,38 @@ import {
...
 import Link from 'components/Link/Link'
-import { convertCollection } from 'convert'
 import useTranslation from 'next-translate/useTranslation'
 import numbro from 'numbro'
 import { FC } from 'react'
 
-type Props = {
-  collection: ReturnType<typeof convertCollection>
+export type Props = {
+  collection: {
```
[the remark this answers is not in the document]
ismailToyran: I will do this one on a new PR after merging this.

components/Bid/AcceptModal.tsx
```diff
@@ -35,7 +35,7 @@ export type Props = {
   onClose: () => void
   bid: {
     id: string
-    availableQuantity: BigNumber
+    availableQuantity: string
```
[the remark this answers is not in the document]
ismailToyran: This is the only one I've questions with.
Can't we change this one on backend to return `BigNumberish`?
I've used backend return types for these props. If I swap to use `Pick<Asset,  'availableQuantity' | ...>` as you've suggested before, these will return `string` instead. So this will not be the fix we should aim.
What do you think?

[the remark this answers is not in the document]
antho1404: We won't be able to change it in the backend as the value is returned in json and we cannot send the `BigNumber` object.
That said, we can change the types (in typescript) we received with Apollo in the `config.scalars` in the `codegen.yml` and replace the `Uint256`/`Int256` into `BigNumberish` (the -ish is important so it matches string like the data from the api)

PR #458, opened 2023-10-23
Fix issue with video stream

antho1404:
Fix issue with some video using application/octet-stream

This can be tested with the collection on uaan https://uaan.liteflow.com/collection/1890/0xf643d2f8975fb87a15e1fa0184c7e2e04eed266f


ismailToyran (approved):
Looking good 👍

PR #461, opened 2023-10-25
Cart system UX improvements

ismailToyran:
Dependant on #460 

**TODO:**
- [x] Add locales


ismailToyran:
@antho1404 could you provide me some feedback on my last commit please?
Added some TODO's for the last missing parts for the drawer.
Hope you like it 😉

components/Navbar/CartDrawerListItem.tsx
```diff
@@ -0,0 +1,142 @@
...
+            colorScheme="gray"
+            variant="outline"
+            onClick={() => removeItem(cartItem.id)}
+          >
+            Remove
+          </Button>
+        ) : // TODO: we could also add increase/decrease quantity buttons here
+        // but could be better to do it in the cart page
+        // user might increase quantity to something that will not be available after some time
+        // so in that case we should show a warning, but there won't be enough space in the drawer
```
antho1404, line 124: We should retrieve the available quantity to limit how much they can set and in the case they put a quantity that is not available anymore yes indeed a message with maybe a small button to update or remove the item would be great

components/CartContext.tsx
```diff
@@ -25,6 +25,10 @@ const CartContext: FC<PropsWithChildren> = (props) => {
     [items],
   )
 
+  // TODO: add fetch when we add first item to the list so we get the chainId for that item
+  // and share that chainId with context so we can block adding items from different chains
+  // no need to do this multiple times, just once when we add first item and clear it when we clear cart
+
```
antho1404, line 31: Hum I'm afraid this will add quite a lot of complexity, maybe more than actually supporting the cart for multichain, maybe let's brainstorm that, for now we can move ahead without this check

components/Sales/Direct/CardFooter.tsx
```diff
@@ -72,26 +73,43 @@ const SaleDirectCardFooter: FC<Props> = ({
...
+      </Flex>
+      {showButton && (
+        <>
+          <Divider orientation="vertical" />
+          <AddToCartButton
+            offerId={saleId}
+            isDisabled={isOwner}
+            borderRadius="none"
+            // add isDisabled prop when the item on the carts are on different chains
+            // also add a tooltip to explain why
```
antho1404, line 107: Similar to another comment, it might be more complicated than actually supporting the multichain, let's keep it like that for now

ismailToyran:
@antho1404 I've added steps as we've discussed, could you have a look and maybe implement approval part on step 2?
Not sure if my implementation was too much and there is an easier way.

components/Cart/CartDrawer.tsx
```diff
@@ -0,0 +1,273 @@
...
+  const { CHAINS } = useEnvironment()
+  const { clearCart, items } = useCart()
+  const [step, setStep] = useState<
+    'selection' | 'transaction' | 'success' | 'error'
+  >('selection')
+  const formValues = useForm<FormData>()
+  const selectedChainId = formValues.watch('chainId')
+  const currencies = formValues.watch('currencies')
+
+  const [fetch, { data }] = useFetchCartItemsLazyQuery({
```
antho1404: Why not use the query directly and not the lazyquery? I like the idea of optimizing the query only when needed but not sure that required yet.

components/Cart/CartDrawer.tsx
```diff
@@ -0,0 +1,273 @@
...
+          </List>
+        )
+      case 'transaction':
+        invariant(cartItems, 'Cart items must be defined')
+        invariant(selectedChain, 'Selected chain must be defined')
+        return (
+          <CartTransactionStep
+            cartItems={cartItems.filter(
+              (cartItem) => cartItem.asset.chainId === selectedChain.id,
+            )}
```
antho1404: Could be good to have this as a variable in a memo with something like `currentCartItems`

components/Cart/CartDrawer.tsx
```diff
@@ -0,0 +1,273 @@
...
+                  as={Link}
+                  href="/explore?offers=fixed"
+                  variant="outline"
+                  mt={3}
+                >
+                  Explore NFTs
+                </Button>
+              </VStack>
+            )}
+          </List>
```
antho1404: Can we maybe accept `cartItems` as optional and handle all this logic in the `CartSelectionStep` that would be cleaner

components/Cart/Step/Selection.tsx
```diff
@@ -0,0 +1,49 @@
...
+  })
+  const chainId = watch('chainId')
+  return (
+    <VStack alignItems="flex-start" width="full" {...getRootProps()}>
+      {chains.map((chain, i) => {
+        const radio = getRadioProps({ value: chain.id })
+        return (
+          <CartRadio key={i} {...radio} isChecked={chain.id === chainId}>
+            <HStack gap={1}>
+              <Checkbox
```
antho1404, line 80: This should be a radio and not checkbox as we can select only one at the time

components/Cart/Step/Selection.tsx
```diff
@@ -0,0 +1,49 @@
...
+                _checked={{ backgroundColor: 'transparent' }}
+                _hover={{ backgroundColor: 'transparent' }}
+              />
+              <Text variant="subtitle2">{chain.name}</Text>
+            </HStack>
+            {cartItems
+              .filter((item) => item.asset.chainId === chain.id)
+              .map((cartItem) => (
+                <CartDrawerListItem key={cartItem.id} cartItem={cartItem} />
+              ))}
```
antho1404, line 92: Would be nicer to have that as part of another component so we can memo the filter. Definitely not an issue now as I don't believe we will have carts with 100 of items

components/Cart/Step/Transaction.tsx
```diff
@@ -0,0 +1,89 @@
...
+import Image from '../../Image/Image'
+import List, { ListItem } from '../../List/List'
+import Price from '../../Price/Price'
+
+type Props = {
+  cartItems: NonNullable<FetchCartItemsQuery['offerOpenSales']>['nodes']
+  chain: { id: number; name: string; image: string }
+}
+
+const CartTransactionStep: FC<Props> = ({ cartItems, chain }) => {
```
antho1404, line 17: I will most likely rework this component on another PR to integrate the business logic, we might do some simplification. All good for now

antho1404 (commented):
Really good, I pushed a few commit for code quality and simplification, feel free to revert if that's not good, a few feedback/questions left but nothing blocker

components/Cart/CartDrawer.tsx
antho1404: I used the direct query https://github.com/liteflow-labs/starter-kit/pull/461/commits/e6989883d32fcb55849199e7a063500ad8f3bd81

components/Cart/CartDrawer.tsx
antho1404: https://github.com/liteflow-labs/starter-kit/pull/461/commits/8fc89ffba22c37e527ef6ac3649ef62f65deb95e

components/Cart/CartDrawer.tsx
antho1404: [c0ff876](https://github.com/liteflow-labs/starter-kit/pull/461/commits/c0ff8768d5772e0ec8591a264713d16e30279ca8)

components/Sales/Direct/CardFooter.tsx
ismailToyran: [Remove some TODO's](https://github.com/liteflow-labs/starter-kit/pull/461/commits/bde5eacc05e9cdc5df54f6a3589c92f067cfa06a)

components/CartContext.tsx
ismailToyran: [Remove some TODO's](https://github.com/liteflow-labs/starter-kit/pull/461/commits/bde5eacc05e9cdc5df54f6a3589c92f067cfa06a)

components/Navbar/CartDrawerListItem.tsx
ismailToyran: [Remove some TODO's](https://github.com/liteflow-labs/starter-kit/pull/461/commits/bde5eacc05e9cdc5df54f6a3589c92f067cfa06a)

components/Cart/Step/Selection.tsx
ismailToyran: Fixed with [Change checkbox to radio](https://github.com/liteflow-labs/starter-kit/pull/461/commits/3a573d275c35fc04d5bfa4c00915bd427bfc01a3)

components/Cart/CartDrawer.tsx
ismailToyran: Only downside of this is, we will fetch all the cart items on load regardless if the user uses cart feature or not.
Might not be big problem, but imagine them adding maybe 50-100 of items?
I'm okay to keep this one like this.

components/Cart/Step/Transaction.tsx
ismailToyran: I'm okay to improve these components on other PR's as well if needed later on 👍

components/Cart/Step/Selection.tsx
ismailToyran: Let's move on and see if this is needed. This part might still be changed with your implementations later on.

components/Cart/CartDrawer.tsx
antho1404: I don't think this will happen anytime soon, and if it does, we may need to rethink the UI anyway.
Let's keep the code simple for now

PR #466, opened 2023-10-30
Supply number added to card for erc1155

ismailToyran:
### Project organization
- Related [trello card](https://trello.com/c/ulljZGbh/1105-feedback-g2a-adding-supply-number-to-erc1155-nfts)

### Description
Added supply number to the NFTs that belongs to ERC1155 (card view on hover)

### Checklist

### DIFF
--- a/components/Token/Card.tsx
+++ b/components/Token/Card.tsx
@@ -1,5 +1,6 @@
 import {
   AspectRatio,
+  Badge,
   Box,
   Flex,
   Heading,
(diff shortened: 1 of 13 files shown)


components/Token/Card.tsx
```diff
@@ -208,25 +213,49 @@ const TokenCard: FC<Props> = ({
...
+              pointerEvents="none"
+              bg="white"
+              color="gray.500"
+            >
+              {t('asset.detail.supply', {
+                supply: numbro(asset.quantity).format({
+                  average: true,
+                  roundingFunction: (num) =>
+                    num > 1000 ? Math.round(num) : num,
+                }),
```
antho1404, line 254: Why this custom rounding function? I think the default should work just fine, and this `Math.round` will just round the decimals like 1341.34 > 1341 which will never happen as we have integers for supply

```suggestion
```

ismailToyran: Right now, if the number is higher than 499, it shows 1k without this check.

antho1404: My point was more on the fact that, in our case, the `Math.round(num) = num`, so this condition doesn't make sense and we could just do `roundingFunction: (num) => num` that should be similar or maybe I am missing something

ismailToyran: Hmm it seems to work like that. I'm confused though, why with `average=true` they don't use this `roundingFunction: (num) => num` as a default. Rounding an item 500 to 1k seems wrong to me.
Anyway thanks for this one, [Better rounding implementation](https://github.com/liteflow-labs/starter-kit/pull/466/commits/a0d65027ebab7a3c9417f073f571e7486451604a) should fix this

PR #470, opened 2023-11-06
SEO implementation to pagination

ismailToyran:
### Project organization
- Related [trello card](https://trello.com/c/5ywfU6h2/1104-feedback-g2a-seo-upgrade-for-pagination-no-links)

### Description
Currently pages with pagination does not contain links, thus search engine robots are not able to reach subsequent subpages.
This PR adds links to the pagination component, so it will result in better SEO for subpages.

### Checklist

### DIFF
--- a/components/Pagination/Pagination.tsx
+++ b/components/Pagination/Pagination.tsx
@@ -11,14 +11,16 @@ import {
 } from '@chakra-ui/react'
 import { IoChevronBackSharp } from '@react-icons/all-files/io5/IoChevronBackSharp'
 import { IoChevronForward } from '@react-icons/all-files/io5/IoChevronForward'
+import Link from 'components/Link/Link'
 import useTranslation from 'next-translate/useTranslation'
+import { useRouter } from 'next/router'
 import { JSX } from 'react'
 
 export type IProp = {
   page: number
   hasNextPage: boolean | undefined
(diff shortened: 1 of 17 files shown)


components/Pagination/Pagination.tsx
```diff
@@ -83,54 +86,66 @@ export default function Pagination({
         gap={4}
       >
         <IconButton
+          as={onPageChange ? undefined : Link}
+          href={{ pathname, query: { ...query, page: page - 1 } }}
```
antho1404, line 90: This will not work when we have the `onPageChange` Button does not have href
I think it would be great to simply get rid of the `onPageChange` if we can and put this state directly in the url with a search params like `?page=xx` this way we have the pagination always with links

ismailToyran: Only place where we don't want url to be changed is owners modal on ERC1155 token detail page.
My implementation right now works even without `onPageChange` feel free to look at `components/Token/Owners/Modal.tsx` 

If you have a better idea for this use case, I will be up for a change. I don't have any other idea for this usage on owners modal.

ismailToyran: [Add undefined for cases](https://github.com/liteflow-labs/starter-kit/pull/470/commits/6d1720d19df2f5dd914f6e3896021cce59967dde)

PR #473, opened 2023-11-10
Optimize loading images on TokenCard component

ismailToyran:
### Project organization
- Related [trello card](https://trello.com/c/R1tnUerH/1106-feedback-g2a-optimize-of-loading-images)

### Description
From now on, TokenCard component will show only image field. Animation will not be used for thumbnail. Animation will still be active on Featured NFT section on homepage and NFT detail page. This is for the NFT cards that is shown on explore, user page etc.

### Checklist


antho1404 (approved):
That does work
We can merge this, ideally we should remove that from the queries as well but that's not critical for now

Issue #363, opened 2023-08-10
Implement pagination on queries with hardcoded limit


NicolasMahe:
A few queries didn't have any limit implemented so I added an hardcoded limit of 100 to them in PR https://github.com/liteflow-labs/starter-kit/pull/362 to limit possible API related issues but without a pagination.

So we now need to implement a pagination system on those queries when needed.
I added the comment `# TODO: implement pagination` to the codebase to find those queries easily.

NicolasMahe:
@ismailToyran could you check what other queries have hardcoded pagination. If that's only the collection's trait value I think we can close this issue for now.

PR #477, opened 2023-11-20
Collection and trait filter improvement

ismailToyran:
### Project organization

- Part of https://github.com/liteflow-labs/starter-kit/issues/363

### Description
- Add infinite loading for collection filter
- Add infinite loading for trait filter
- Add proper way of searching for filters

### Checklist


hooks/useFetchCollectionTraits.gql
```diff
@@ -1,11 +1,23 @@
-query FetchCollectionTraits($chainId: Int!, $address: Address!) {
+query FetchCollectionTraits(
+  $chainId: Int!
+  $address: Address!
+  $filter: CollectionTraitFilter!
+  $limit: Int!
+  $cursor: Cursor
```
NicolasMahe, line 6: Please use offset instead of cursor, even if you will need to keep track of the current offset in a state.
There are some edge cases when cursor is not working correctly.

components/Filter/FilterBy/Trait.tsx
```diff
@@ -11,54 +14,96 @@ import {
...
-      chainId: (collection && collection.chainId) || 0,
+      address: collection.address,
+      chainId: collection.chainId,
+      cursor: null,
+      limit: PAGINATION_LIMIT,
+      filter: {
+        type: {
+          includesInsensitive: filterResult.propertySearch || '',
+        } as StringFilter,
+      } as CollectionTraitFilter,
```
NicolasMahe, line 62: Could you try to remove the fallback value on `filterResult.propertySearch` and not pass any filter, something like this:
```suggestion
      filter: filterResult.propertySearch ? {
        type: {
          includesInsensitive: filterResult.propertySearch,
        } as StringFilter,
      } : {} as CollectionTraitFilter,
```
This will better in term of SQL query executed by the backend

hooks/useFetchCollectionTraits.gql
ismailToyran: [use offset instead of cursor](https://github.com/liteflow-labs/starter-kit/pull/477/commits/a454cc2898ef55b309f51040e7f7af70e59a736a)

NicolasMahe (approved):
Pagination and search are working great 👍 

I noticed an issue when searching a trait and then selecting the trait. The issue exists before this PR and this PR actually improve it a bit. I will create a dedicated issue: https://github.com/liteflow-labs/starter-kit/issues/479

PR #482, opened 2023-11-22
Add cart button to the detail page and offer modal

antho1404:
Add missing "add to cart" button by creating a dedicated "checkout" button

Screenshot 2023-11-22 at 18 57 56
Screenshot 2023-11-22 at 18 58 08

@ismailToyran could you check that there is no other places where we need to add that

### DIFF
--- a/components/Button/CheckoutButton.tsx
+++ b/components/Button/CheckoutButton.tsx
@@ -1,0 +1,52 @@
+import { Button, ButtonGroup, ButtonProps, Divider } from '@chakra-ui/react'
+import { JSX, PropsWithChildren, useMemo } from 'react'
+import useAccount from '../../hooks/useAccount'
+import { isSameAddress } from '../../utils'
+import Link from '../Link/Link'
+import AddToCartButton from './AddToCart'
+
+type Props = Omit<ButtonProps, 'href'> & {
+  offer: {
+    id: string
+    maker: { address: string }
+  }
+}
+
+export default function CheckoutButton({
+  offer,
+  children,
+  ...props
+}: PropsWithChildren<Props>): JSX.Element {
+  const { address } = useAccount()
+
+  const isDisabled = useMemo(
+    () => (address ? isSameAddress(address, offer.maker.address) : false),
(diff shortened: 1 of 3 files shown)


components/Button/CheckoutButton.tsx
```diff
@@ -0,0 +1,52 @@
...
+  )
+
+  return (
+    <ButtonGroup isAttached width="full" isDisabled={isDisabled}>
+      <Button
+        {...props}
+        width="full"
+        as={Link}
+        isDisabled={isDisabled}
+        href={isDisabled ? '#' : `/checkout/${offer.id}`}
```
ismailToyran, line 34: Do we really want to disable link?
For both components it's being checked outside of this component to show correct one.
I believe isDisabled is not needed in this case.

components/Button/CheckoutButton.tsx
```diff
@@ -0,0 +1,52 @@
...
+        borderRightRadius={address ? 0 : undefined}
+      >
+        {children}
+      </Button>
+      {address && (
+        <>
+          <Divider width="1px" />
+          <AddToCartButton
```
ismailToyran, line 42:
```suggestion
        <>
          <Divider orientation="vertical" />
          <AddToCartButton
```

ismailToyran (approved):
I couldn't find any other place for direct purchase either
These 2 were the only ones left. Nice catch 👍

components/Button/CheckoutButton.tsx
antho1404: I'm good to simplify it [8941ad0](https://github.com/liteflow-labs/starter-kit/pull/482/commits/8941ad0239f010ad7b70cee24bc8946c426bda55)

components/Button/CheckoutButton.tsx
antho1404: [8941ad0](https://github.com/liteflow-labs/starter-kit/pull/482/commits/8941ad0239f010ad7b70cee24bc8946c426bda55)

PR #499, opened 2023-12-04
Fix and improve cart callback

antho1404:
Fix issue when navigating due to a function not memoized, leading to adding infinite callback making the call stack quite huge...
Also cleaned up the callbacks when the hook is destroyed

### DIFF
--- a/components/CartContext.tsx
+++ b/components/CartContext.tsx
@@ -17,7 +17,14 @@ const CartContext: FC<PropsWithChildren> = (props) => {
   >([])
 
   const registerOnCheckout = useCallback((callback: () => void) => {
-    setOnCheckoutCallbacks((prev) => [...prev, callback])
+    setOnCheckoutCallbacks((prev) => [
+      ...prev.filter((x) => x !== callback),
+      callback,
+    ])
+  }, [])
+
+  const unregisterOnCheckout = useCallback((callback: () => void) => {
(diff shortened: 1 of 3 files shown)


pages/index.tsx
```diff
@@ -16,7 +16,8 @@ type Props = {
 }
 const HomePage: NextPage<Props> = ({ now }) => {
   const [date, setDate] = useState(new Date(now))
-  useCart({ onCheckout: () => setDate(new Date()) })
```
antho1404, line 19: This anonymous function was recreated at each render as the object is recreated at each render as well

components/CartContext.tsx
```diff
@@ -17,7 +17,14 @@ const CartContext: FC<PropsWithChildren> = (props) => {
   >([])
 
   const registerOnCheckout = useCallback((callback: () => void) => {
-    setOnCheckoutCallbacks((prev) => [...prev, callback])
+    setOnCheckoutCallbacks((prev) => [
+      ...prev.filter((x) => x !== callback),
+      callback,
+    ])
```
NicolasMahe, line 23: Could be better to not call `setOnCheckoutCallbacks` if the callback already exist. Similar comment apply to `unregisterOnCheckout`.
I'm worried that calling `setOnCheckoutCallbacks` even with the same elements will trigger a re-render of this component that will trigger a cascade of re-render, especially that this is a context and loaded very early in the components tree.

PR #515, opened 2023-12-21
Allow multiple offer listing

antho1404:
Simpler version of https://github.com/liteflow-labs/starter-kit/pull/513

Only disable the artificial restriction, we had to not create a new offer if one existed, and to avoid having 2 big call to actions, cleaned up the cancel section to be less dominant

### DIFF
--- a/components/Sales/Direct/Info.tsx
+++ b/components/Sales/Direct/Info.tsx
@@ -1,15 +1,5 @@
-import {
-  Box,
-  Flex,
-  Heading,
-  Icon,
-  Stack,
-  Text,
-  useDisclosure,
-  useToast,
-} from '@chakra-ui/react'
+import { Stack, Text, useDisclosure, useToast } from '@chakra-ui/react'
 import { CancelOfferStep, useCancelOffer } from '@liteflow/react'
-import { BiBadgeCheck } from '@react-icons/all-files/bi/BiBadgeCheck'
 import { HiArrowNarrowRight } from '@react-icons/all-files/hi/HiArrowNarrowRight'
 import useAccount from 'hooks/useAccount'
 import useSigner from 'hooks/useSigner'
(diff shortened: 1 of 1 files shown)


components/Sales/Direct/Info.tsx
```diff
@@ -166,17 +119,14 @@ const SaleDirectInfo: FC<Props> = ({
 
   const create = useMemo(() => {
     if (!isOwner) return null
-    // Hide the create button if we already show the cancel button to avoid having multiple call to action
-    // This can be disabled if we want to also let the user create new offers
-    if (cancel) return null
```
antho1404, line 171: This is where we were hiding the button that is not necessary for multiple listing

components/Sales/Direct/Info.tsx
```diff
@@ -82,52 +71,16 @@ const SaleDirectInfo: FC<Props> = ({
...
-            <Text
-              as={Price}
-              color="brand.black"
-              ml={2}
-              fontWeight="semibold"
-              amount={currentAccountFirstSale.unitPrice}
-              currency={currentAccountFirstSale.currency}
-            />
-          </Heading>
-        </Flex>
```
antho1404, line 130: This whole section was just to make it nicer when we have a single action but now only the button is fine

components/Sales/Direct/Info.tsx
```diff
@@ -82,52 +71,16 @@ const SaleDirectInfo: FC<Props> = ({
   const cancel = useMemo(() => {
     // Only display cancel when there is a single offer owned by the current account
     if (!address) return null
-    const currentAccountFirstSale = sales.find((x) =>
+    const currentAccountSales = sales.filter((x) =>
       isSameAddress(x.maker.address, address),
     )
+    if (currentAccountSales.length !== 1) return null
```
antho1404, line 77: Still keep the cancel button only if there is only one offer from the current user

antho1404:
closing and focusing on fixing https://github.com/liteflow-labs/starter-kit/pull/513

PR #520, opened 2024-01-08
Add currency count indicator to SaleOpenSummary component

ismailToyran:
### Description
There shouldn't be more than 5 currency to show on token details anymore.
Instead we show count indicator to reduce the space it takes.

This is an example with many currencies and how it should look now,

image

### Checklist


antho1404:
Are you sure about the calculation? If this screenshot is taken from Hodooi, I count 23 currencies, we then display the first 5 so it should be 23-5=18 and not 19

PR #522, opened 2024-01-10
Add compatibility with Neon EVM

NicolasMahe:
### Description

Add compatibility with Neon EVM.

Waiting for the green light to merge it.

### Checklist


environment.ts
```diff
@@ -259,7 +259,38 @@ const getEnvironment = async (): Promise<Environment> => {
             url: 'https://pegasus.lightlink.io',
           },
         },
-      } as Chain,
+      },
+      {
+        id: 245022926,
```
NicolasMahe, line 264: This config if available in a more recent version of wagmi but I updating it seems to also require to update rainbow kit so I just copy/paste the config.

environment.ts
```diff
@@ -259,7 +259,38 @@ const getEnvironment = async (): Promise<Environment> => {
...
+        network: 'neonDevnet',
+        name: 'Neon EVM DevNet',
```
NicolasMahe, line 266: Need to add the icon of this chain in `/public/chains/245022926.svg`

NicolasMahe: also need to add icon for the mainnet chain: `/public/chains/245022934.svg`

NicolasMahe: icons added

environment.ts
NicolasMahe: I pushed a PR to remove the need to add every chains in `connectors.ts` here: https://github.com/liteflow-labs/starter-kit/pull/537

Issue #489, opened 2023-11-29
Cart system feedbacks


NicolasMahe:
Let's centralize here the cart system feedbacks to not block PR https://github.com/liteflow-labs/starter-kit/pull/471

NicolasMahe:
- [ ] Add a go back button when displaying an error. It's really annoying to close the cart and reopen again
<img width="367" alt="Screenshot 2023-11-29 at 15 48 32" src="https://github.com/liteflow-labs/starter-kit/assets/5823445/13f9a3cf-cbaa-47b1-843b-27f7b4d3f42b">

- [ ] Replace the network icon on the button "Continue with [icon]" by the network name
<img width="367" alt="Screenshot 2023-11-29 at 15 50 21" src="https://github.com/liteflow-labs/starter-kit/assets/5823445/1b7c9577-7257-46b9-8125-f5e7511c39a7">

- [ ] The hover and click area on the nft in the cart doesn't match. I need to click on the picture or texts to be redirected to the nft.
<img width="367" alt="Screenshot 2023-11-29 at 15 53 10" src="https://github.com/liteflow-labs/starter-kit/assets/5823445/6085f6ae-871f-4f65-9521-e19084fc4dd9">

- [ ] Add html hint / tooltip on the add to cart and remove from cart button. As a user I'm not sure what action the icons are going to trigger, specially the `-` button:
<img width="367" alt="Screenshot 2023-11-29 at 15 54 27" src="https://github.com/liteflow-labs/starter-kit/assets/5823445/ecb9d69c-23f2-4a58-8678-adb431be0f7b">

- [ ] Open the cart when I add a NFT. It's not super obvious how to open the cart.

- [ ] No more modal with the transaction status? Specially when doing the approval, it takes time and I don't have much info.

- [ ] There is no selection of quantity for ERC1155. It purchase one edition only. Is that normal? Do we plan to add it later?

EmmanuelDrouin:
Feedback based on Nicolas proposal:
- I agree on all feedback if I do not comment those

> The hover and click area on the nft in the cart doesn't match. I need to click on the picture or texts to be redirected to the nft.

I think we should only give the ability here to get redirected to the item and not the collection. I feel like this will be more efficient for the users. This is the behavrio on Opensea and Rarible, you can only get redirected to the item, no link to the collection

> Open the cart when I add a NFT. It's not super obvious how to open the cart.

[...]

PR #526, opened 2024-01-16
Cart system feedbacks

ismailToyran:
### Project organization
- Closes https://github.com/liteflow-labs/starter-kit/issues/489

### Description
Improvements on UI-UX regarding the comments on the issue.

### Improvement
Might need to update `@liteflow/react` to allow better import of `BatchPurchaseStep`
We should be using `import { BatchPurchaseStep } from '@liteflow/react'` instead of `import { BatchPurchaseStep } from '@liteflow/react/dist/useBatchPurchase'`

### Checklist


antho1404 (approved):
One small issue left when paying, when clicking, the button doesn't have a loader and it feels during the time that the wallet open that I didn't click.

Also, when the payment is in progress there is no label in the button, it looks buggy
Screenshot 2024-01-23 at 09 58 28


Other than that all good. Feel free to merge and work on that on another PR of fix here directly

PR #527, opened 2024-01-18
Update dependencies

antho1404:
Update all dependencies and fix new formatting/linting/types

Notable lib updates:
- Next
- wagmi/rainbowkit/viem
- next-translate
- typescript/eslint/prettier


components/MarkdownViewer.tsx
```diff
@@ -73,7 +73,7 @@ const MarkdownViewer: FC<MarkdownPreviewProps> = ({ source, ...props }) => {
             h4: (props) => <Text {...props} />,
             h5: (props) => <Text {...props} />,
             h6: (props) => <Text {...props} />,
-            code: (props) => <Text {...props} />,
+            code: (props) => <Text {...(props as any)} />, // TODO: fix this type
```
antho1404, line 76: Not sure how to fix this type but haven't spent much time on it

connectors.ts
```diff
@@ -1,4 +1,4 @@
-import { MagicConnectConnector } from '@everipedia/wagmi-magic-connector'
+import { UniversalWalletConnector } from '@magiclabs/wagmi-connector'
```
antho1404, line 1: @everipedia/wagmi-magic-connector library has been deprecated because moved to the official magiclabs repo which is a good news as we bet on the right library that is now maintained by the team of magic

next.config.js
```diff
@@ -1,4 +1,4 @@
-const nextTranslate = require('next-translate')
+const nextTranslate = require('next-translate-plugin')
```
antho1404, line 1: The breaking change of next-translate that changed their lib structure for apparently great optimization

ismailToyran: Nice, the last time I've tried to update this library, it was causing errors on the Chat page.
Seems to be working without changing anything right now. 🎉

ismailToyran (approved):
Looking great,
Didn't face any errors (tested on local + development preview)
Really happy with the `next/translate` update

PR #532, opened 2024-01-23
Display deleted data

NicolasMahe:
### Description

- display history related to deleted data in AssetHistory query
- display notifications of deleted collection and trades. I didn't remove any links because neither point to the token page, they all point to user activity pages
- display trades related to deleted assets and currencies in pages `users/trades/purchased`, and `users/trades/sold`
  - disable links to asset page
- display offers related to deleted assets and currencies in pages `users/bids/placed`, `users/bids/received`, and `users/offers`
  - disable links to asset page and button to accept or create new bid
- display 404:
  - if asset is deleted on pages tokens
  - if collection is deleted on collection pages
  - if asset or currency is deleted on checkout page

### Checklist

### DIFF
--- a/components/History/HistoryList.gql
+++ b/components/History/HistoryList.gql
@@ -19,6 +19,8 @@ query FetchAssetHistory(
       orderBy: DATE_DESC
       first: $limit
       offset: $offset
+      includeWhenAssetDeleted: YES
+      includeWhenCurrencyDeleted: YES
     ) {
       pageInfo {
         hasNextPage
--- a/layouts/navbar.gql
+++ b/layouts/navbar.gql
@@ -3,7 +3,11 @@ query NavbarAccount($account: Address!, $lastNotification: Datetime!) {
     address
     name
     image
-    notifications(filter: { createdAt: { greaterThan: $lastNotification } }) {
+    notifications(
+      filter: { createdAt: { greaterThan: $lastNotification } }
+      includeWhenCollectionDeleted: YES
+      includeWhenTradeDeleted: YES
+    ) {
       totalCount
     }
   }
--- a/pages/notification.gql
+++ b/pages/notification.gql
@@ -17,6 +17,8 @@ query GetNotifications($address: Address!, $cursor: Cursor) {
     first: 12
     orderBy: CREATED_AT_DESC
     after: $cursor
+    includeWhenCollectionDeleted: YES
+    includeWhenTradeDeleted: YES
   ) {
     pageInfo {
       hasNextPage
--- a/pages/users/[id]/bids/placed.gql
+++ b/pages/users/[id]/bids/placed.gql
@@ -14,6 +14,8 @@ query FetchUserBidsPlaced(
       signature: { isNull: false }
       makerAddress: { equalTo: $address }
     }
+    includeWhenAssetDeleted: YES
+    includeWhenCurrencyDeleted: YES
   ) {
     pageInfo {
       hasNextPage
@@ -49,6 +51,7 @@ query FetchUserBidsPlaced(
         name
         image
         imageMimetype
+        deletedAt
       }
     }
   }
--- a/pages/users/[id]/bids/placed.tsx
+++ b/pages/users/[id]/bids/placed.tsx
@@ -163,7 +163,11 @@ const BidPlacedPage: NextPage = () => {
                       <Td>
                         <Flex
                           as={Link}
-                          href={`/tokens/${item.asset.id}`}
+                          href={
+                            item.asset.deletedAt
+                              ? undefined // no link if asset is deleted
+                              : `/tokens/${item.asset.id}`
+                          }
                           gap={3}
                         >
                           <Image
@@ -239,7 +243,11 @@ const BidPlacedPage: NextPage = () => {
                             ) : (
                               <Button
                                 as={Link}
-                                href={`/tokens/${item.asset.id}/bid`}
+                                href={
+                                  item.asset.deletedAt
+                                    ? undefined // no link if asset is deleted
+                                    : `/tokens/${item.asset.id}/bid`
+                                }
                                 variant="outline"
                                 colorScheme="gray"
                               >
--- a/pages/users/[id]/bids/received.gql
+++ b/pages/users/[id]/bids/received.gql
@@ -17,6 +17,8 @@ query FetchUserBidsReceived(
       makerAddress: { notEqualTo: $address }
       asset: { ownerships: { some: { ownerAddress: { equalTo: $address } } } }
     }
+    includeWhenAssetDeleted: YES
+    includeWhenCurrencyDeleted: YES
   ) {
     pageInfo {
       hasNextPage
@@ -43,6 +45,7 @@ query FetchUserBidsReceived(
         id
         decimals
         symbol
+        deletedAt
       }
       asset {
         id
@@ -52,6 +55,7 @@ query FetchUserBidsReceived(
         name
         image
         imageMimetype
+        deletedAt
       }
     }
   }
--- a/pages/users/[id]/bids/received.tsx
+++ b/pages/users/[id]/bids/received.tsx
@@ -172,7 +172,11 @@ const BidReceivedPage: NextPage<Props> = ({ now }) => {
                       <Td>
                         <Flex
                           as={Link}
-                          href={`/tokens/${item.asset.id}`}
+                          href={
+                            item.asset.deletedAt
+                              ? undefined // no link if asset is deleted
+                              : `/tokens/${item.asset.id}`
+                          }
                           gap={3}
                         >
                           <Image
@@ -223,25 +227,27 @@ const BidReceivedPage: NextPage<Props> = ({ now }) => {
                       </Td>
                       <Td>{dateFromNow(item.createdAt)}</Td>
                       <Td isNumeric>
-                        {ownerLoggedIn && (
-                          <AcceptOfferButton
-                            offer={item}
-                            title={t('user.bid-received.accept.title')}
-                            variant="outline"
-                            colorScheme="gray"
-                            onAccepted={onAccepted}
-                            onError={(e) =>
-                              toast({
-                                status: 'error',
-                                title: formatError(e),
-                              })
-                            }
-                          >
-                            <Text as="span" isTruncated>
-                              {t('user.bid-received.actions.accept')}
-                            </Text>
-                          </AcceptOfferButton>
-                        )}
+                        {ownerLoggedIn &&
+                          !item.asset.deletedAt && // only display accept button if asset and currency are not deleted
+                          !item.currency.deletedAt && (
+                            <AcceptOfferButton
+                              offer={item}
+                              title={t('user.bid-received.accept.title')}
+                              variant="outline"
+                              colorScheme="gray"
+                              onAccepted={onAccepted}
+                              onError={(e) =>
+                                toast({
+                                  status: 'error',
+                                  title: formatError(e),
+                                })
+                              }
+                            >
+                              <Text as="span" isTruncated>
+                                {t('user.bid-received.actions.accept')}
+                              </Text>
+                            </AcceptOfferButton>
+                          )}
                       </Td>
                     </Tr>
                   ))}
--- a/pages/users/[id]/offers.tsx
+++ b/pages/users/[id]/offers.tsx
@@ -134,7 +134,11 @@ const FixedPricePage: NextPage = () => {
                       <Td>
                         <Flex
                           as={Link}
-                          href={`/tokens/${item.asset.id}`}
+                          href={
+                            item.asset.deletedAt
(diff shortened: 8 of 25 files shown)


pages/users/[id]/bids/placed.tsx
```diff
@@ -163,7 +163,11 @@ const BidPlacedPage: NextPage = () => {
                       <Td>
                         <Flex
                           as={Link}
-                          href={`/tokens/${item.asset.id}`}
+                          href={
+                            item.asset.deletedAt
+                              ? 'LINK TO DISABLE' // no link if asset is deleted
```
Comment (NicolasMahe) [inline pages/users/[id]/bids/placed.tsx:168]: @ismailToyran Please I need help here to remove the link

pages/users/[id]/bids/received.tsx
```diff
@@ -172,7 +172,11 @@ const BidReceivedPage: NextPage<Props> = ({ now }) => {
                       <Td>
                         <Flex
                           as={Link}
-                          href={`/tokens/${item.asset.id}`}
+                          href={
+                            item.asset.deletedAt
+                              ? 'LINK TO DISABLE' // no link if asset is deleted
```
Comment (NicolasMahe) [inline pages/users/[id]/bids/received.tsx:177]: @ismailToyran Please I need help here to remove the link

pages/users/[id]/offers.tsx
```diff
@@ -134,7 +134,11 @@ const FixedPricePage: NextPage = () => {
                       <Td>
                         <Flex
                           as={Link}
-                          href={`/tokens/${item.asset.id}`}
+                          href={
+                            item.asset.deletedAt
+                              ? 'LINK TO DISABLE' // no link if asset is deleted
```
Comment (NicolasMahe) [inline pages/users/[id]/offers.tsx:139]: @ismailToyran Please I need help here to remove the link

pages/users/[id]/trades/purchased.tsx
```diff
@@ -173,7 +173,11 @@ const TradePurchasedPage: NextPage = () => {
                         {item.asset ? (
                           <Flex
                             as={Link}
-                            href={`/tokens/${item.asset.id}`}
+                            href={
+                              item.asset.deletedAt
+                                ? 'LINK TO DISABLE' // no link if asset is deleted
```
Comment (NicolasMahe) [inline pages/users/[id]/trades/purchased.tsx:178]: @ismailToyran Please I need help here to remove the link

pages/users/[id]/trades/sold.tsx
```diff
@@ -165,7 +165,11 @@ const TradeSoldPage: NextPage = () => {
                         {item.asset ? (
                           <Flex
                             as={Link}
-                            href={`/tokens/${item.asset.id}`}
+                            href={
+                              item.asset.deletedAt
+                                ? 'LINK TO DISABLE' // no link if asset is deleted
```
Comment (NicolasMahe) [inline pages/users/[id]/trades/sold.tsx:170]: @ismailToyran Please I need help here to remove the link

pages/users/[id]/bids/received.gql
```diff
@@ -17,6 +17,8 @@ query FetchUserBidsReceived(
       makerAddress: { notEqualTo: $address }
       asset: { ownerships: { some: { ownerAddress: { equalTo: $address } } } }
     }
+    includeWhenAssetDeleted: YES
+    includeWhenCurrencyDeleted: YES
```
Comment (NicolasMahe) [inline pages/users/[id]/bids/received.gql:21]: @antho1404  Does it make sense to display bid that the user cannot accept anymore?

pages/users/[id]/bids/placed.tsx
Comment (ismailToyran) [reply pages/users/[id]/bids/placed.tsx:168]: Only thing I could suggest is, not to touch href attribute but add a new one to disable touching the links
It should be `pointerEvents="none"`.

If this doesn't work out, even though there will be some duplication, we should do something like `asset.deletedAt ? <Flex /> : <Button as={Link} href={...} />`

Comment (NicolasMahe) [reply pages/users/[id]/bids/placed.tsx]: Thanks! The `pointerEvents="none"` seems to do the trick.

Comment (antho1404) [reply pages/users/[id]/bids/placed.tsx]: Can we instead have a condition on the `Link` component with something like `<Link if={!item.asset.deletedAt}>...`

That would be a lot easier and we already used this pattern in the dashboard

components/History/HistoryList.gql
```diff
@@ -19,6 +19,8 @@ query FetchAssetHistory(
       orderBy: DATE_DESC
       first: $limit
       offset: $offset
+      includeWhenAssetDeleted: YES
+      includeWhenCurrencyDeleted: YES
```
antho1404, line 23: Can we have the currency set to `YES` by default in the API so we don't need to care about this case everywhere as discussed the currency should never impact the display of data

components/History/HistoryList.gql
```diff
@@ -19,6 +19,8 @@ query FetchAssetHistory(
       orderBy: DATE_DESC
       first: $limit
       offset: $offset
+      includeWhenAssetDeleted: YES
```
antho1404, line 22: I'm not sure this is needed, by design we shouldn't display this asset as it has been deleted so we won't display its activity

layouts/navbar.gql
```diff
@@ -3,7 +3,11 @@ query NavbarAccount($account: Address!, $lastNotification: Datetime!) {
     address
     name
     image
-    notifications(filter: { createdAt: { greaterThan: $lastNotification } }) {
+    notifications(
+      filter: { createdAt: { greaterThan: $lastNotification } }
+      includeWhenCollectionDeleted: YES
```
antho1404, line 8: Can we somehow force on the API to set these ones to YES by default for the notifications

pages/notification.gql
```diff
@@ -17,6 +17,8 @@ query GetNotifications($address: Address!, $cursor: Cursor) {
     first: 12
     orderBy: CREATED_AT_DESC
     after: $cursor
+    includeWhenCollectionDeleted: YES
+    includeWhenTradeDeleted: YES
```
antho1404, line 21: Similar comment about the possibility to move that to the API

pages/users/[id]/bids/placed.gql
```diff
@@ -14,6 +14,8 @@ query FetchUserBidsPlaced(
       signature: { isNull: false }
       makerAddress: { equalTo: $address }
     }
+    includeWhenAssetDeleted: YES
+    includeWhenCurrencyDeleted: YES
```
Comment (antho1404) [inline pages/users/[id]/bids/placed.gql:18]: Same comment about the currency, the asset is good here I think

pages/users/[id]/bids/placed.tsx
```diff
@@ -164,6 +164,9 @@ const BidPlacedPage: NextPage = () => {
                         <Flex
                           as={Link}
                           href={`/tokens/${item.asset.id}`}
+                          pointerEvents={
+                            item.asset.deletedAt ? 'none' : undefined // disable link if asset is deleted
+                          }
```
Comment (antho1404) [inline pages/users/[id]/bids/placed.tsx:169]: This will be bad for SEO as the link still exists and will redirect to a 404 we need to disable the whole link not just the pointer

pages/users/[id]/bids/placed.tsx
```diff
@@ -237,16 +240,18 @@ const BidPlacedPage: NextPage = () => {
...
-                                colorScheme="gray"
-                              >
-                                <Text as="span" isTruncated>
-                                  {t('user.bid-placed.actions.new')}
-                                </Text>
-                              </Button>
+                              !item.asset.deletedAt && ( // only display new button if asset is not deleted
+                                <Button
+                                  as={Link}
+                                  href={`/tokens/${item.asset.id}/bid`}
```
Comment (antho1404) [inline pages/users/[id]/bids/placed.tsx:246]: I would suggest something like that

```suggestion
                                <Button
                                  as={Link}
                                  if={!item.asset.deletedAt}
                                  href={`/tokens/${item.asset.id}/bid`}
```

pages/users/[id]/bids/received.gql
Comment (antho1404) [reply pages/users/[id]/bids/received.gql:21]: Yes but we should prevent the user to accept it. If we don't display it that would be weird as the user will get the notification

pages/users/[id]/trades/purchased.gql
```diff
@@ -9,6 +9,9 @@ query FetchUserTradePurchased(
     first: $limit
     offset: $offset
     filter: { buyerAddress: { equalTo: $address } }
+    includeWhenAssetDeleted: YES
+    includeWhenCollectionDeleted: YES
```
Comment (antho1404) [inline pages/users/[id]/trades/purchased.gql:13]: I don't think this one is useful as it should be covered by the asset already
```suggestion
```

Comment (antho1404) [reply pages/users/[id]/trades/purchased.gql:13]: Also for the trades, didn't we say that we will display the trades of deleted assets only if it was purchased from the platform so basically if an offer exists

pages/users/[id]/trades/sold.gql
```diff
@@ -9,6 +9,9 @@ query FetchUserTradeSold(
     first: $limit
     offset: $offset
     filter: { sellerAddress: { equalTo: $address } }
+    includeWhenAssetDeleted: YES
+    includeWhenCollectionDeleted: YES
```
Comment (antho1404) [inline pages/users/[id]/trades/sold.gql:13]:
```suggestion
```

pages/users/[id]/bids/received.gql
Comment (NicolasMahe) [reply pages/users/[id]/bids/received.gql:21]: Alright so all good

components/History/HistoryList.gql
NicolasMahe: No I cannot change the default only for currency, it will change the default for all of those properties.
Also the currency DOES impact the display of data, for example offers and bids related to a deleted currency on the token page, or any token card.

components/History/HistoryList.gql
NicolasMahe: that's right but for simplicity, I implemented it as all or nothing to avoid headache and case by case logic.

layouts/navbar.gql
NicolasMahe: No I cannot as well.

pages/users/[id]/trades/purchased.gql
Comment (NicolasMahe) [reply pages/users/[id]/trades/purchased.gql:13]: > I don't think this one is useful as it should be covered by the asset already

See comment https://github.com/liteflow-labs/starter-kit/pull/532#discussion_r1464527484

pages/users/[id]/trades/sold.gql
Comment (NicolasMahe) [reply pages/users/[id]/trades/sold.gql:13]: See comment https://github.com/liteflow-labs/starter-kit/pull/532#discussion_r1464527484

pages/users/[id]/bids/placed.tsx
Comment (NicolasMahe) [reply pages/users/[id]/bids/placed.tsx]: done in https://github.com/liteflow-labs/starter-kit/pull/532/commits/4207b88b2b43fc3eea62829a01054684422c895e
I use the same implementation as the admin

pages/users/[id]/bids/placed.tsx
Comment (NicolasMahe) [reply pages/users/[id]/bids/placed.tsx]: done in https://github.com/liteflow-labs/starter-kit/pull/532/commits/4207b88b2b43fc3eea62829a01054684422c895e
I use the same implementation as the admin

pages/users/[id]/bids/placed.tsx
Comment (NicolasMahe) [reply pages/users/[id]/bids/placed.tsx:244]: done in https://github.com/liteflow-labs/starter-kit/pull/532/commits/4207b88b2b43fc3eea62829a01054684422c895e
I use the same implementation as the admin

pages/users/[id]/trades/purchased.gql
Comment (NicolasMahe) [reply pages/users/[id]/trades/purchased.gql:13]: Filter on offers added in https://github.com/liteflow-labs/starter-kit/pull/532/commits/caebf7d2965904e10f9b953b61b3d030c8bbe39a

layouts/navbar.gql
antho1404: Hum this one will be bad but maybe something that we need to think about to fix on the new API

pages/users/[id]/trades/purchased.gql
Comment (antho1404) [reply pages/users/[id]/trades/purchased.gql:13]: We talked about this part for the dashboard and we also said that the starter kit should never have more data than the dashboard

pages/users/[id]/trades/purchased.gql
```diff
@@ -8,7 +8,7 @@ query FetchUserTradePurchased(
     orderBy: $orderBy
     first: $limit
     offset: $offset
-    filter: { buyerAddress: { equalTo: $address } }
+    filter: { buyerAddress: { equalTo: $address }, offerExists: true }
```
Comment (antho1404) [inline pages/users/[id]/trades/purchased.gql:11]: This one is not enough, we need all the trades not deleted or the trades delete where the offer exists.
Here that will only display the trades where the offer exists removing all the trades from external platforms

Comment (NicolasMahe) [reply pages/users/[id]/trades/purchased.gql]: Alright I now understand what you meant in the other comment!
Change done in https://github.com/liteflow-labs/starter-kit/pull/532/commits/e3a891b1326a373a22f7c06473146b9a23e265e5.
I had to update the API to expose the `deletedAt` filters.

antho1404 (approved):

PR #540, opened 2024-01-25
Remove chain from query filter in collection page

NicolasMahe:
### Description

Remove chain from query filter in collection page to create a lighter query.

### Checklist


pages/collection/[chainId]/[id]/index.tsx
```diff
@@ -111,6 +111,7 @@ const CollectionPage: FC<Props> = ({ now }) => {
         currency,
         collectionSearch, // exclude from query
         propertySearch, // exclude from query
+        chains, // exclude from filter and query, already set by from the collection
         ...otherFilters
       } = filter
```
Comment (NicolasMahe) [inline pages/collection/[chainId]/[id]/index.tsx:116]: @antho1404 any reason the `otherFilters` is passed to `cleanData` which end up in the URL?
Wouldn't it make more sense to explicitly extract what we want to filter from `filter`? eg:
```ts
      const {
        traits,
        currency,
        collection, // Doesn't seem to be needed
        maxPrice,
        minPrice,
        offers,
        search, // Doesn't seem to be needed
      } = filter
      const cleanData = removeEmptyFromObject({
        ...Object.keys(query).reduce((acc, value) => {
          if (value.startsWith('trait')) return acc
          return { ...acc, [value]: query[value] }
        }, {}),
        collection,
        maxPrice,
        minPrice,
        offers,
        search,
        currency: currency?.id,
        decimals: currency?.decimals,
        page: 1,
        ...traits.reduce(
          (acc, { type, values }) => ({
            ...acc,
            [`traits[${type}]`]: values,
          }),
          {},
        ),
      })
```

Comment (antho1404) [reply pages/collection/[chainId]/[id]/index.tsx:116]: This is to make sure we can get all the other filters by default, so if we add a filter, it will be added automatically, and we don't have to edit all the places (and forget most of them), but indeed that leads to some extra params like here passed which I still believe is better than having a filter that doesn't work.

antho1404 (approved):

PR #542, opened 2024-01-26
Feature / Homepage banners added

ismailToyran:
### Project organization
- Related [trello card](https://trello.com/c/OiuRHxd8/1186-marketplace-template-hero-banner-customization)

### Description
By using dashboard, dashboard user can now add banner(s) to the homepage.

### Checklist


EmmanuelDrouin:
One feedback on the implementation, can we have a skeleton effect while loading the background image?
- Currently, if the image is a bit heavy, it is pretty slow to load on my side (images from 3MB are slow)
- in the mean-time the image is loading, only the text and button are displayed. If you have a white background on your platform and set your text color to white, then it looks buggy as nothing shows up for a few seconds on the screen

Adding a skeleton loader for the component will improve the UX and remove the bug feeling for the end users

EmmanuelDrouin:
One additional point we could change maybe right now for consistency is to rename the section from “Starter kit configuration” to “Marketplace template configuration” as we stop using the “starter kit” wording.

ismailToyran:
@antho1404 have any idea how we can improve the loading of the image?
As we upload them on admin, we get `IPFS` link. And that's why it takes blank screen for a while.
After first time use this shouldn't be issue.

Also we don't have this loader on token cards either, same loading strategy. But we have background color for this case.
So I've added background + `priority` tag to the banner images.

components/HomeSection/Banner.tsx
```diff
@@ -0,0 +1,65 @@
...
+import Link from '../../components/Link/Link'
+import MarkdownViewer from '../../components/MarkdownViewer'
+import useEnvironment from '../../hooks/useEnvironment'
+import Image from '../Image/Image'
+import Slider from '../Slider/Slider'
+
+const BannerHomeSection = () => {
+  const { HOME_BANNERS } = useEnvironment()
+
+  if (HOME_BANNERS.length === 0) return null
```
antho1404, line 11: Could you also add a check if this exists as this will probably break for existing platforms which didn't update their settings yet

antho1404 (commented):
I think the priority + background should be enough.
Just one thing to fix to ensure existing platforms will not crash and then we are good to merge

components/HomeSection/Banner.tsx
ismailToyran: Changed to `if (!HOME_BANNERS.length) return null` like rest of the HomeSection components
These shouldn't be problem as we return empty array if there is no data coming from the dashboard.

```ts
    // Home page configuration
    HOME_BANNERS: metadata.HOME_BANNERS || [],
    FEATURED_TOKEN: metadata.FEATURED_TOKEN || [],
    HOME_COLLECTIONS: metadata.HOME_COLLECTIONS || [],
    HOME_USERS: metadata.HOME_USERS || [],
    HOME_TOKENS: metadata.HOME_TOKENS || [],
```

antho1404: Indeed all good then, I missed this default value

antho1404 (approved):

PR #543, opened 2024-01-29
Fix filter price component

ismailToyran:
### Project organization
- Related [trello card](https://trello.com/c/m2OtMFUm/1193-issue-default-displayed-currency-in-price-range-filter-is-actually-not-really-the-default)

### Description
Fix filter price component.

### Checklist

### DIFF
--- a/components/Filter/FilterBy/Price.tsx
+++ b/components/Filter/FilterBy/Price.tsx
@@ -17,7 +17,7 @@ import {
   Text,
 } from '@chakra-ui/react'
 import useTranslation from 'next-translate/useTranslation'
-import { FC, useMemo } from 'react'
+import { FC, useCallback, useEffect, useMemo } from 'react'
 import { UseFormReturn } from 'react-hook-form'
 import { useFetchCurrenciesQuery } from '../../../graphql'
 import { Filter } from '../../../hooks/useAssetFilterFromQuery'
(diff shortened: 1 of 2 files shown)


ismailToyran:
This fixes the issue for select currency but there is still an issue with the prices after reset.
Couldn't manage to fix it. Spent some much time. Could create an issue just for that case.
Also for the order. I've removed it for this select. We shouldn't really order the `choices` inside the select component.
That way default value and the first item of the select usually will not be the same. We should order on query only. What do you think @antho1404 ? Otherwise we need to sort on many places to also have the default value right.

components/Filter/FilterBy/Price.tsx
```diff
@@ -58,13 +58,31 @@ const FilterByPrice: FC<Props> = ({
...
+  useEffect(() => {
+    if (currency) {
+      setValue('currency', { id: currency.id, decimals: currency.decimals })
+    }
+    if (filterResult.minPrice) {
+      setValue('minPrice', filterResult.minPrice)
+    }
+    if (filterResult.maxPrice) {
+      setValue('maxPrice', filterResult.maxPrice)
+    }
```
antho1404, line 75: I don't understand the goal of that, `filterResult` is already the data from the form (that we get from `watch` so for `minPrice` and `maxPrice` we are setting the value when the value changed but the value already changed because we are in the useEffect of filterResult

components/Filter/FilterBy/Price.tsx
```diff
@@ -58,13 +58,31 @@ const FilterByPrice: FC<Props> = ({
...
-    )
+    return filterResult.currency
+      ? currencies.find((x) => x.id === filterResult.currency?.id)
+      : currencies[0]
   }, [currencies, filterResult.currency])
 
+  useEffect(() => {
+    if (currency) {
+      setValue('currency', { id: currency.id, decimals: currency.decimals })
+    }
```
antho1404, line 69: Similar for this one, this is indirectly similar because `currency` is a variation of the `filterResult` already

components/Select/Select.tsx
```diff
@@ -64,7 +64,10 @@ const Select = <T,>({
   ...props
 }: IProps<T>): ReactElement => {
   const selectedChoice = useMemo(
-    () => choices.find((x) => x.value === value),
+    () =>
+      typeof value === 'string'
+        ? choices.find((x) => x.value === value)
+        : choices.find((x: any) => x.value.id === value?.id),
```
antho1404, line 70: I really don't like this one, we are doing a dedicated select just for the pricing I would really recommend against that.

antho1404 (commented):
> Also for the order. I've removed it for this select. We shouldn't really order the choices inside the select component.
I kind of disagree on that, the order we put on the select shouldn't impact anything, indeed we could sort in the query but ordering in a select shouldn't be an issue. The issue here is the selection of the default value so we need to make sure we select the first currency after ordering if the select is ordering to have something consistent. Ordering on the query tho would solve all in one indeed.

Also I noticed something that could be improved and fix some issues:
- Having the currency object in the select makes things complicated and is not necessary, we only need the id and then where needed get the decimals based on the id of the currency (and not get it from the URL otherwise we can have a currency with 18 decimals but the user changes that to 10 decimals and breaks everything)
- The query to fetch the currencies `useFetchCurrenciesQuery` could be done in the `FilterAsset` (and also connected to to the chain selected instead of having to filter it locally) and currencies should be pass to the `FilterByPrice` component

I think with that, it will remove some weird hack we currently have that are mainly the source of these issues

ismailToyran:
Can be reopened later on.

PR #546, opened 2024-01-30
Fix / ERC1155 token card fixed price sale cases

### LINKED ISSUE #483: Allow purchase of 1155 when we are owner
If we own items of an ERC1155, the button displayed in the card displays the wrong text and the cart button is disabled. We should be able to purchase as long as we have an offer available for us https://github.com/liteflow-labs/starter-kit/pull/476/commits/cf445d82992763c8bf08b173e1dd0fc4522778ec

ismailToyran:
### Project organization

- Closes #483 
- Related [trello card](https://trello.com/c/5ueJPkeR/1196-issue-allow-purchase-of-1155-when-we-are-owner)

### Description
- Fix wrong text case for ERC1155 token card
- Fix disabled cart add button case for ERC1155 on token card
- Fix not logged in user case for token card

### Further improvement
is tracked on https://github.com/liteflow-labs/starter-kit/issues/529
Need to brainstorm

### Checklist

### DIFF
--- a/components/Sales/Direct/CardFooter.tsx
+++ b/components/Sales/Direct/CardFooter.tsx
@@ -33,6 +33,8 @@ const SaleDirectCardFooter: FC<Props> = ({
   const { t } = useTranslation('components')
   const { address } = useAccount()
 
+  // TODO: we should have a modal if there is more than one sale like we have on detail page
+  // issue is tracked on https://github.com/liteflow-labs/starter-kit/issues/529 for this modal improvement
   const isOwner = useMemo(
     () => (address ? isSameAddress(sale.maker.address, address) : false),
     [address, sale.maker],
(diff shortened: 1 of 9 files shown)


components/Sales/Direct/CardFooter.tsx
```diff
@@ -94,7 +106,7 @@ const SaleDirectCardFooter: FC<Props> = ({
             : chip}
         </Text>
       </Flex>
-      {showButton && (
+      {showButton && address && (
```
antho1404, line 109: It's not ideal for this one; that fixes a bug, but maybe we can think of a better system where we don't hide the button or just disable it.
For now, we can keep it like that; I will create an issue.

antho1404 (approved):

PR #554, opened 2024-02-01
Flatten notifications in `NavbarAccount` query

NicolasMahe:
### Description

Flatten notifications in `NavbarAccount` query

### Checklist

### DIFF
--- a/layouts/navbar.gql
+++ b/layouts/navbar.gql
@@ -3,12 +3,15 @@ query NavbarAccount($account: Address!, $lastNotification: Datetime!) {
     address
     name
     image
-    notifications(
-      filter: { createdAt: { greaterThan: $lastNotification } }
-      includeWhenCollectionDeleted: YES
-      includeWhenTradeDeleted: YES
-    ) {
-      totalCount
+  }
+  notifications(
+    filter: {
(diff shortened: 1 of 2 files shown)


layouts/navbar.gql
```diff
@@ -1,14 +1,20 @@
...
   account(address: $account) {
     address
     name
     image
-    notifications(
-      filter: { createdAt: { greaterThan: $lastNotification } }
-      includeWhenCollectionDeleted: YES
-      includeWhenTradeDeleted: YES
-    ) {
-      totalCount
```
NicolasMahe, line 11: @antho1404 would it make sense in the API v1 to have a `account.notificationCount(from: Datetime!)` resolver?
On one side it's would be practical but it would mean notifications are still (loosely) connected to an account.

layouts/navbar.gql
```diff
@@ -1,14 +1,20 @@
...
+  }
+}
+
+query NavbarNotification($account: Address!, $lastNotification: Datetime!) {
```
NicolasMahe, line 9: TODO @NicolasMahe: instead of split the whole query, can just move notifications under account.

PR #566, opened 2024-02-04
Flatten `FetchDrops` query

NicolasMahe:
### Description

Flatten `FetchDrops` query

### How to test

Before: https://dev.uaan.liteflow.com/drops
After: https://nft-test-uaan-git-chore-flatten-fetchdrops-query-liteflow.vercel.app/drops

### Checklist

### DIFF
--- a/components/Drop/DropCard.tsx
+++ b/components/Drop/DropCard.tsx
@@ -9,7 +9,6 @@ import {
   Stack,
   Text,
 } from '@chakra-ui/react'
-import { BigNumber } from '@ethersproject/bignumber'
 import { HiArrowNarrowRight } from '@react-icons/all-files/hi/HiArrowNarrowRight'
 import { HiBadgeCheck } from '@react-icons/all-files/hi/HiBadgeCheck'
 import Trans from 'next-translate/Trans'
@@ -25,21 +24,21 @@ import Link from '../Link/Link'
 import Price from '../Price/Price'
 
 type Props = {
-  collection: {
-    address: string
-    chainId: number
-    cover: string | null
-    image: string | null
-    name: string
-    deployer: {
-      address: string
-      name: string | null
-      verification: {
-        status: AccountVerificationStatus
-      } | null
-    }
-  }
   drop: {
+    collection: {
+      address: string
+      chainId: number
+      cover: string | null
+      image: string | null
+      name: string
+      deployer: {
+        address: string
+        name: string | null
+        verification: {
+          status: AccountVerificationStatus
+        } | null
+      }
+    }
     startDate: Date
     endDate: Date
     unitPrice: string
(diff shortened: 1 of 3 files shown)


components/Drop/DropCard.tsx
```diff
@@ -154,34 +148,34 @@ export default function DropCard({
...
             </Text>
-            {collection.deployer.verification?.status === 'VALIDATED' && (
+            {drop.collection.deployer.verification?.status === 'VALIDATED' && (
               <Icon as={HiBadgeCheck} color="brand.500" h={4} w={4} />
             )}
           </Flex>
 
           <Flex alignItems="center" gap={2}>
-            {totalSupply ? (
+            {drop.supply ? (
```
NicolasMahe, line 168: I replace the collection's drops totalSupply by simply the drop's supply. It will display the current or last drop's supply on the card.

pages/drops.tsx
```diff
@@ -27,50 +25,59 @@ const DropsPage: NextPage<Props> = ({ now }) => {
...
+        .filter(
+          (drop, i, arr) =>
+            arr.findIndex(
+              (x) =>
+                x.collection.chainId === drop.collection.chainId &&
+                x.collection.address === drop.collection.address,
+            ) === i,
+        )
+    )
+  }, [data?.active?.nodes, data?.ended?.nodes])
```
NicolasMahe, line 66: Before of those filter and with the pagination system, the number of drops displayed is not going to be constant.

pages/drops.gql
```diff
@@ -1,110 +1,44 @@
...
   }
-  ended: collections(
-    orderBy: [DROPS_MIN_START_DATE_ASC]
-    filter: {
-      dropsExist: true
-      drops: { every: { endDate: { lessThanOrEqualTo: $now } } }
-    }
+  ended: drops(
+    orderBy: [START_DATE_DESC]
+    filter: { endDate: { lessThanOrEqualTo: $now } }
```
NicolasMahe, line 41: the order of the replaced query was actually weird, an ASC in the top level but a DESC in the children.
Now it's simply a DESC

pages/drops.tsx
antho1404: This will break the whole pagination.
First, for the `endedDrops`, we could have 2 separate queries with the `ended` query that would filter out the list of collections from the active drops.
This won't solve the duplication within the same category, though, and I don't believe we have a `distinct` in the API.

One other way to see that and that I would recommend is to for now simplify this and don't mind about the deduplication of collections

NicolasMahe: I'm all good with duplications of collections if they have multiple drops. I will implement the changes now.

NicolasMahe: Confirmed

PR #573, opened 2024-02-12
Replace filters by conditions in all queries

NicolasMahe:
### Description

Replace filters by conditions in all queries.

As this is a big PR and it uses new conditions that I implemented myself on the backend, @antho1404 @ismailToyran can you check if there is no regression. Also you can familiarize yourself with the new way to filter data using the `condition` object instead of the `filter`.

Only staging backend is updated. I'm waiting for first feedback to deploy the new version of the backend on prod.

### Checklist

### DIFF
--- a/components/Cart/Step/Selection.gql
+++ b/components/Cart/Step/Selection.gql
@@ -1,5 +1,5 @@
 query FetchCartItems($offerIds: [UUID!]!) {
-  offerOpenSales(filter: { id: { in: $offerIds } }) {
+  listings(condition: { ids: $offerIds, status: ACTIVE }) {
     nodes {
       id
       type
--- a/components/Cart/Step/Selection.gql
+++ b/components/Cart/Step/Selection.gql
@@ -1,5 +1,5 @@
 query FetchCartItems($offerIds: [UUID!]!) {
-  listings(condition: { ids: $offerIds, status: ACTIVE }) {
+  listings(condition: { ids: $offerIds, status: [ACTIVE, EXPIRED] }) {
     nodes {
       id
       type
--- a/components/HomeSection/Assets.gql
+++ b/components/HomeSection/Assets.gql
@@ -1,9 +1,5 @@
 query FetchDefaultAssetIds($limit: Int!) {
-  assets(
-    filter: { quantity: { greaterThan: "0" } }
-    orderBy: CREATED_AT_DESC
-    first: $limit
-  ) {
+  assets(orderBy: CREATED_AT_DESC, first: $limit) {
     nodes {
       chainId
       collectionAddress
@@ -12,11 +8,12 @@ query FetchDefaultAssetIds($limit: Int!) {
   }
 }
 
-query FetchAssets($limit: Int!, $filter: [AssetFilter!]!, $address: Address) {
-  assets(
-    filter: { quantity: { greaterThan: "0" }, and: $filter }
-    first: $limit
-  ) {
+query FetchAssets(
+  $limit: Int!
+  $ids: [AssetIdsCondition!]!
+  $address: Address
+) {
+  assets(condition: { ids: $ids }, first: $limit) {
     nodes {
       chainId
       collectionAddress
--- a/hooks/useAssetFilterFromQuery.ts
+++ b/hooks/useAssetFilterFromQuery.ts
@@ -1,21 +1,10 @@
 import { useRouter } from 'next/router'
 import type { ParsedUrlQuery } from 'querystring'
 import invariant from 'ts-invariant'
-import {
-  AssetFilter,
-  AssetToManyOfferFilter,
-  IntFilter,
-  OfferFilter,
-  Uint256Filter,
-} from '../graphql'
+import { AssetCondition, AssetTraitsCondition } from '../graphql'
 import { parseBigNumber } from './useParseBigNumber'
 import useQueryParamMulti from './useQueryParamMulti'
 import useQueryParamSingle from './useQueryParamSingle'
-
-type TraitFilter = {
-  type: string
-  values: string[]
-}
 
 export type Filter = {
   chains: number[]
@@ -28,7 +17,7 @@ export type Filter = {
     id: string
     decimals: number
   } | null
-  traits: TraitFilter[]
+  traits: AssetTraitsCondition[]
   collectionSearch?: string
   propertySearch?: string
 }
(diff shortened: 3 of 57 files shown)


package.json
```diff
@@ -63,7 +63,7 @@
     "@graphql-codegen/typescript-react-apollo": "^4.1.0",
     "@graphql-eslint/eslint-plugin": "^3.20.1",
     "@next/bundle-analyzer": "^14.0.4",
-    "@nft/api-graphql": "^1.0.0-beta.52-prerelease-3",
+    "@nft/api-graphql": "^1.0.0-beta.52-prerelease-4",
```
NicolasMahe, line 66: To update to non-prerelease

components/HomeSection/Assets.gql
```diff
@@ -12,11 +8,12 @@ query FetchDefaultAssetIds($limit: Int!) {
   }
 }
 
-query FetchAssets($limit: Int!, $filter: [AssetFilter!]!, $address: Address) {
-  assets(
-    filter: { quantity: { greaterThan: "0" }, and: $filter }
```
NicolasMahe, line 17: Note for reviewer: the API directly filter out asset with no quantity on the `assets` query

components/Cart/Step/Selection.gql
```diff
@@ -1,5 +1,5 @@
 query FetchCartItems($offerIds: [UUID!]!) {
-  offerOpenSales(filter: { id: { in: $offerIds } }) {
+  listings(condition: { ids: $offerIds, status: ACTIVE }) {
```
NicolasMahe, line 2: `offerOpenSales`, `offerOpenBuys` and `offers` queries are now deprecated and replaced by `listings` and `openOffers`. To get the "open" ones, need to set the condition status to active.

hooks/useSearchCollection.gql
```diff
@@ -1,12 +1,13 @@
...
+  $chainIds: [Int!]
+  $search: String
   $limit: Int!
   $offset: Int!
 ) {
   collections(
     orderBy: TOTAL_VOLUME_DESC
     first: $limit
-    filter: $filter
+    condition: { chainIds: $chainIds, search: $search }
```
NicolasMahe, line 10: Note for reviewer: only the `search` condition can partially match data.
On this `collections` query, it will execute a "contains case-insensitive" against `name`, `description`, `deployerAddress` and `address`.

layouts/navbar.gql
```diff
@@ -5,10 +5,7 @@ query NavbarAccount($account: Address!, $lastNotification: Datetime!) {
     image
   }
   notifications(
-    filter: {
-      accountAddress: { equalTo: $account }
```
NicolasMahe, line 9: Note for reviewer: no need to filter accountAddress, it's done by API.

pages/users/[id]/owned.gql
```diff
@@ -6,7 +6,7 @@ query FetchOwnedAssets(
   $orderBy: [OwnershipsOrderBy!]
 ) {
   owned: ownerships(
-    filter: { ownerAddress: { equalTo: $address }, assetExists: true }
+    condition: { ownerAddress: $address }
```
Comment (NicolasMahe) [inline pages/users/[id]/owned.gql:9]: Note for reviewer: API directly filters out Ownership without asset.

pages/users/[id]/bids/placed.gql
```diff
@@ -4,16 +4,11 @@ query FetchUserBidsPlaced(
...
     orderBy: $orderBy
     first: $limit
     offset: $offset
-    filter: {
-      type: { equalTo: BUY }
-      availableQuantity: { greaterThan: "0" }
-      signature: { isNull: false }
-      makerAddress: { equalTo: $address }
-    }
+    condition: { status: ACTIVE, makerAddress: $address }
```
Comment (antho1404) [inline pages/users/[id]/bids/placed.gql:11]: I have a diff on this one, some of the bids disappeared

https://uaan.liteflow.com/users/0x4b595014f7b45789c3f4e79324ae6d8090a6c8b5/bids/placed
https://nft-test-uaan-h6s0w9t49-liteflow.vercel.app/users/0x4b595014f7b45789c3f4e79324ae6d8090a6c8b5/bids/placed

components/Cart/Step/Selection.gql
antho1404: @ismailToyran were we actually displaying the old one with some disabled states or something like that?

antho1404 (commented):
Pretty much all good, one "issue" on the bids, and I still have the explore page to test more, @ismailToyran please have a look as well I might have missed some parts

pages/users/[id]/bids/placed.gql
Comment (NicolasMahe) [reply pages/users/[id]/bids/placed.gql:11]: Good catch 👍  The `expiredAt` was not set in the filter thus the `status: ACTIVE` is not correct.

Comment (NicolasMahe) [reply pages/users/[id]/bids/placed.gql]: Fixed: https://nft-test-uaan-git-chore-replace-filters-by-conditions-liteflow.vercel.app/users/0x4b595014f7b45789c3f4e79324ae6d8090a6c8b5/bids/placed

Actually I was able to make the condition on `status` compatible with both single value and array of values by implementing the array of values version. GraphQL seems to automatically convert the single value to an array.

components/Cart/Step/Selection.gql
ismailToyran: Yes we have expired case for this one.
```ts
  const isExpired = useMemo(
    () => !dateIsBefore(new Date(), offer.expiredAt),
    [offer.expiredAt],
  )
```
so I believe this should be `status: ['ACTIVE', 'EXPIRED']`?

ismailToyran (approved):
One comment regarding [components/Cart/Step/Selection.gql](https://github.com/liteflow-labs/starter-kit/pull/573/files/3517313beac17a03efef17af8e57eb3b54952395#diff-51944b63c1d705a08734f980bca8702f494623f670caa39118bd3c4f1af0a8b6)

Other than that looks really good. From the queries I've seen, it simplifies things pretty good 💪

hooks/useAssetFilterFromQuery.ts
```diff
@@ -150,25 +46,66 @@ export const extractTraitsFromQuery = (
...
+  }
+
   if (filter.currency) {
-    if (filter.minPrice)
-      queryFilter.push(minPriceFilter(filter.minPrice, filter.currency, now))
-    if (filter.maxPrice)
-      queryFilter.push(maxPriceFilter(filter.maxPrice, filter.currency, now))
+    queryFilter.listings = {
+      currencyId: filter.currency.id,
+      makerAddress: null,
```
antho1404, line 69: Why this `makerAddress: null` I don't think this makes sense as all listings should have a maker

hooks/useAssetFilterFromQuery.ts
```diff
@@ -150,25 +46,66 @@ export const extractTraitsFromQuery = (
...
+    queryFilter.chainId = chainId ? parseInt(chainId, 10) : null
+    queryFilter.collectionAddress = collectionAddress ? collectionAddress : null
+  }
+
   if (filter.currency) {
-    if (filter.minPrice)
-      queryFilter.push(minPriceFilter(filter.minPrice, filter.currency, now))
-    if (filter.maxPrice)
-      queryFilter.push(maxPriceFilter(filter.maxPrice, filter.currency, now))
+    queryFilter.listings = {
```
antho1404, line 67: The price filter should also work for open offers right now it only applies to listings.
So if I want to filter with `has offer` and price min 0.001 I can't

hooks/useAssetFilterFromQuery.ts
NicolasMahe: It's because the codegen use in the starter-kit forces to set every value. I cannot omit it or pass undefined.
I will try something with a type cast to get around this.

hooks/useAssetFilterFromQuery.ts
antho1404: Hum indeed, that was a bug then, all good to merge this like that but can you create an issue to track this bug, if we select "has offer" we should be able to filter the price for the offers (or we get rid of this filter)

hooks/useAssetFilterFromQuery.ts
NicolasMahe: null removed and cast added in https://github.com/liteflow-labs/starter-kit/pull/573/commits/d0b80aaced8dc0fe75b370c86db79316189fee9e

hooks/useAssetFilterFromQuery.ts
NicolasMahe: Issue created: https://github.com/liteflow-labs/starter-kit/issues/578

components/Cart/Step/Selection.gql
NicolasMahe: updated in https://github.com/liteflow-labs/starter-kit/pull/573/commits/a216b94ebef1234d90320fb828364b6a5b0ded2d

